PR #5, opened 2015-05-28
Adding index to addEntry/removeEntry

yasserf:


src/record/list.js
```diff
@@ -209,4 +217,27 @@ List.prototype._applyUpdate = function( message ) {
...
+ *
+ * @param {Number} index
+ *
+ * @private
+ * @returns {Number}
+ */
+List.prototype._hasIndex = function( index ) {
+	var hasIndex = false;
+	var entries = this.getEntries();
+	if( typeof index !== 'undefined' ) {
```
WolframHempel: Very minor, but if( index !== undefined) is a bit quicker since it doesn't create two strings

src/record/list.js
```diff
@@ -209,4 +217,27 @@ List.prototype._applyUpdate = function( message ) {
...
+		if( isNaN( index ) ) {
+			throw new Error( 'Index must be a number' );
+		}
+		if( index > entries.length || index < 0 ) {
```
WolframHempel: should be if( index >= entries.length ...

PR #13, opened 2015-06-30
+ banner for minified file

tomByrer:
Reasoning: files often concocted (use: [jsDelivr CDN](https://github.com/jsdelivr/jsdelivr#load-multiple-files-with-single-http-request), so you'll lose file ID/name & copyright.

Should work, but untested.


tomByrer:
> license is misspelt

Darn living in EU & un-learning American English ;)
cheers

PR #28, opened 2015-10-09
Added add-/move-/remove-entry event to list

WolframHempel:


src/record/list.js
```diff
@@ -214,7 +226,9 @@ List.prototype._applyUpdate = function( message ) {
 		message.data[ 2 ] = '[]';
```
yasserf: i know this isn't related, but this means you could listen to any record as a list and be told its empty rather than it not being a list.

WolframHempel: Yup, you can subscribe to lists as records at the moment as well. Once we have proper support for record options we should store a flag in the meta data to differenciate between records and lists

yasserf: Yeah, I meant if I listen to a record called 'user' which is { name: 'John' } via the list API what it will do is rewrite the content with '[]'

That being said, it's a different issue from this PR and not exactly a priority.

yasserf:
Looks good, should we raise an issue for website doc?

PR #34, opened 2015-12-05
Fixes #27 connecting after voluntary disconnect

### LINKED ISSUE #27: Force disconnect/login a client that is connected
We have an Android application that is showing our web-application in a web view (we're using Crosswalk's webview, https://crosswalk-project.org/). When the application is loosing its focus (i.e. the user minimizes the application) we need to disconnect our socket, and when the application is regaining its focus we need to login with the socket again. Or is this the wrong way to do it?

We know when the application is loosing focus and when it's regaining focus.

We've tried to close the socket (with `deepstreamClient.close()`), and then login with it again. However, this gives us the error: `this client's connection was closed`, which is wanted since we actually closed the socket.

What we would need is a `deepstreamClient.disconnect()` or some kind of re-login on the current socket.

Is this possible to implement?
I can make a PR if I could get some hints where do add this functionality so you guys don't have to implement this =)

yasserf:


src/message/connection.js
```diff
@@ -137,6 +135,24 @@ Connection.prototype.close = function() {
 };
 
 /**
+ *
```
WolframHempel: Would you mind adding a sentence describing what this method does

yasserf:
Added doc

PR #44, opened 2015-12-27
Bugs/record issues from specs

yasserf:


yasserf:
Didn't add tests since they were exposed and caught by the gherkin tests. Wonder if Travis has a build trigger..

PR #48, opened 2015-12-31
Feature/resubscribe events after disconnect

yasserf:
This still needs tests (  they are covered by the spec tests ). Wanted a code review before writing the tests tho.

Pretty simply put, I extracted logic from record and applied it to events and the listen object, so now when we lose a connection and regain it we get all the event subscriptions and listen queries back.

Need it with RPCs too, but not touching that yet =P

Also, this based of the bug/event branch since that has all the current fixes for the spec test. Actualy changes are quite minal in comparison.


src/event/event-handler.js
```diff
@@ -120,25 +128,44 @@ EventHandler.prototype.unlisten = function( pattern ) {
  * @returns {void}
  */
 EventHandler.prototype._$handle = function( message ) {
-	var name;
-	
+	var name = message.action === C.ACTIONS.ACK ? message.data[ 1 ] : message.data[ 0 ];
```
WolframHempel: Hehe, if we're going for hardcore compact, why not 
```javascript
var name = message.data[ message.action === C.ACTIONS.ACK ? 1 : 0 ];
```

yasserf: Good point haha

src/event/event-handler.js
```diff
@@ -120,25 +128,44 @@ EventHandler.prototype.unlisten = function( pattern ) {
...
+
 	if( message.action === C.ACTIONS.EVENT ) {
 		if( message.data && message.data.length === 2 ) {
-			this._emitter.emit( message.data[ 0 ], messageParser.convertTyped( message.data[ 1 ] ) );
+			this._emitter.emit( name, messageParser.convertTyped( message.data[ 1 ] ) );
```
WolframHempel: convertTyped needs client as second argument too

yasserf: good catch, that was't caught by tests so will add one for that as well.

src/event/event-handler.js
```diff
@@ -120,25 +128,44 @@ EventHandler.prototype.unlisten = function( pattern ) {
...
+		return;
+	}
+	
+	if( message.action === C.ACTIONS.ERROR ) {
+		message.processedError = true;
+		this._client._$onError( C.TOPIC.EVENT, message.data[ 0 ], message.data[ 1 ] );
+		return;
+	}
+
+	this._client._$onError( C.TOPIC.RECORD, C.EVENT.UNSOLICITED_MESSAGE, name );
```
WolframHempel: topic should be EVENT

src/message/message-parser.js
```diff
@@ -60,10 +60,12 @@ MessageParser.prototype.convertTyped = function( value, client ) {
 	}
 	
 	if( type === C.TYPES.OBJECT ) {
-		try{
```
WolframHempel: Good catch with `processedError`

src/record/record.js
```diff
@@ -249,17 +250,17 @@ Record.prototype.whenReady = function( callback ) {
  */
 Record.prototype._$onMessage = function( message ) {
 	if( message.action === C.ACTIONS.READ ) {
-		this._clearTimeouts();
+		clearTimeout( this._readTimeout );
```
WolframHempel: Wouldn't we want to clear readAck Timeout here as well (the ACK isn't mandatory)

yasserf: readAck is cleared as a normal ack from server as part of the handle ack workflow

R|A|S|recordName+

src/utils/ack-timeout-registry.js
```diff
@@ -32,12 +32,16 @@ EventEmitter( AckTimeoutRegistry.prototype );
...
  */
-AckTimeoutRegistry.prototype.add = function( name ) {
-	if( this._register[ name ] ) {
-		throw new Error( 'Timeout ' + name + ' is already registered' );
+AckTimeoutRegistry.prototype.add = function( name, action ) {
+	var uniqueName = action ? action + name : name;
+	
+	if( this._register[ uniqueName ] ) {
+		this.clear( {
+			data: [ action, name ]
```
WolframHempel: Shouldn't it be 
```javascript
this.clear({
    action: action,
    data: [ name ]
})
```
?

yasserf: its an ACK reply so action is ACK and the real action is data[0]

R|A|S|recordName

action: A
data: [ S, recordName ]

WolframHempel: True that :-)

yasserf: Does this mean I passed the test? (a) =P

src/record/record.js
WolframHempel: Yup, but only getting a read without preceding ACK should be a valid case. Otherwise we'll get the data, the record is ready and then the ACK timeout triggers and throws an error

src/utils/reconnection-notifier.js
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Makes sure that all functionality is resubscribed on reconnect. Subscription is called
+ * when the connection drops - which seems counterintuitive, but in fact just means
+ * that the re-subscription message will be added to the queue of messages that
+ * need re-sending as soon as the connection is re-established.
+ * 
+ * Resubscribe logic should only occur once per connection loss
+ *
+ * @param {Client} client          The deepstream client
+ * @param {Function} reconnect     Function to call to allow resubscribing
```
WolframHempel: (Nitpicking) could we call this method (and associate this._reconnect) "resubscribe" please

src/record/record.js
yasserf: True, although ack timeout is whether or not we are subscribed to changes and response timeout is the time to get the data from the server. I get what you mean by if we get the data read it already means subscribe should pass, but since packets are sent serially if we don't get a subscribe ack before it means something went wrong too.

Is it possible to ever receive a response timeout without a subscribe ack?

src/utils/reconnection-notifier.js
```diff
@@ -0,0 +1,53 @@
...
+ * Check whenever the connection state changes if it is in reconnecting to resubscribe
+ * @private
+ * @returns {void}
+ */
+ ReconnectionNotifier.prototype._handleConnectionStateChanges = function() {
+    var state = this._client.getConnectionState();
+        
+    if( state === C.CONNECTION_STATE.RECONNECTING && this._isReconnecting === false ) {
+        this._isReconnecting = true;
+        this._reconnect();
```
WolframHempel: Shouldn't reconnect (actually resubscribe, see above) be called once the connection is open again, as opposed to when its lost? (So move this to line 50?)

src/utils/reconnection-notifier.js
yasserf: Yep, makes sense!

src/utils/reconnection-notifier.js
WolframHempel: (The reason it worked anyway I guess is due to our fabulous offline mode that queues messages)

yasserf: I was thinking/playing with that too! Coz like this way it gets sent before the auth which doesn't make sense either.

Should be sent after authenticate right?

yasserf: which is OPEN*

sorry, daft moment

test-unit/unit/utils/ack-timeout-registrySpec.js
```diff
@@ -13,28 +13,29 @@ describe( 'webrtc sends correct messages', function(){
...
+	it( 'adding an entry twice does not throw error', function(){
+		registry.add( 'testA', 'S' );
+		expect( mockClient.lastError ).toBe( null );
 	});
 
-	it( 'receives an ACK message', function( done ){
+	it( 'receives an ACK message clears timeout', function( done ){
 		mockClient.lastError = null;
 
 		registry.clear({
```
WolframHempel: Shall we add some test for the setAction feature?

yasserf:
@WolframHempel I added a bit more logic to get the tests passing correctly, so won't be merging now, but it's in what I would call 'The perfect merging state' *win*

PR #70, opened 2016-02-12
Support for data in login response

Svenskunganka:
This adds support to receive data in login callback, e.g:

```javascript
client.login({ username: 'username', password: 'password' }, function(success, event, error, data) {
  console.log(data) // { property: 'value' }
})
```

See: [deepstreamIO/deepstream.io#89](https://github.com/deepstreamIO/deepstream.io/pull/89) & [deepstreamIO/deepstream.io#88](https://github.com/deepstreamIO/deepstream.io/issues/88)

@WolframHempel If you look through the tests you'll see that we've got a really ugly implementation due to backwards-compatibility, and we should clean it up as soon as we get ready for a breaking release.

### DIFF
--- a/src/message/connection.js
+++ b/src/message/connection.js
@@ -37,7 +37,7 @@ var Connection = function( client, url, options ) {
 
 /**
  * Returns the current connection state.
- * (One of constants.CONNECTION_STATE) 
+ * (One of constants.CONNECTION_STATE)
  *
  * @public
  * @returns {String} connectionState
@@ -68,7 +68,7 @@ Connection.prototype.authenticate = function( authParams, callback ) {
 		this._client.once( C.EVENT.CONNECTION_STATE_CHANGED, this.authenticate.bind( authParams, callback ) );
 		return;
 	}
-	
+
 	this._authParams = authParams;
 	this._authCallback = callback;
 
@@ -111,7 +111,7 @@ Connection.prototype.send = function( message ) {
 		this._currentMessageResetTimeout = utils.nextTick( this._resetCurrentMessageCount.bind( this ) );
 	}
 
-	if( this._state === C.CONNECTION_STATE.OPEN && 
+	if( this._state === C.CONNECTION_STATE.OPEN &&
 		this._queuedMessages.length < this._options.maxMessagesPerPacket &&
 		this._currentPacketMessageCount < this._options.maxMessagesPerPacket ) {
 		this._sendQueuedMessages();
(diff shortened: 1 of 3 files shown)


Svenskunganka:
Failing test due to [deepstreamIO/deepstream.io#89](https://github.com/deepstreamIO/deepstream.io/pull/89) isn't merged yet

yasserf:
We discussed this earlier and both wanted to change the API around, This would make it a major release tho, since it's a breaking change.

svenskunganka [6:50 PM] 
@yasserf: this is really ugly: `expect( authCallback ).toHaveBeenCalledWith( true, undefined, undefined, undefined );` :disappointed:

yasserf [6:51 PM] 
true, ideally it would be:
`expect( authCallback ).toHaveBeenCalledWith( true, {}, ); //undefined, undefined );

I assumed the data returned would always be JSON rather than typed as well, same as login takes a json map as auth parameters, but flexibility does rule.

WolframHempel:
That brings us to the question of how religious we want to be about semantic versioning? I think that going for a poorer API choice to comply with a naming concept might not be the best decision, especially for a project as young as ds.

So, should we do semantic versioning by the books?

### Pro
- Yes, because it's an established standard and is used by third party services like Greenkeeper to determine the potential effects of an update

### Against
- You loose the other, more established meaning behind versions. 0.x = beta, >1,x production, >2.x totally new version
- We loose the ability to communicate fundamental updates, (e.g. Angular 2 being completely different from Angular 1)
- SemVer (read: proper Semver) is not as established as we think. Node itself (where Semver originated) remained in 0.x.x for the largest part of its lifespan.

Personally I don't feel strongly either way. Deepstream is used in production by numerous projects, so a 1.x.x version is certainly warranted, but I don't think the world will fall apart if we do another 0.x.x release with this change either.

Logically, it isn't a breaking change anyways. Those users that ignored checking the isSuccesful boolean and instead went for

```javascript
ds.login( data, function( isSuccess, error, errorMsg ) {
    if( !error ) {
      // success
    }
});
```
Won't be using the now supported data parameter, so will be fine. I'd just really like to avoid the undefined undefined thing.

@Svenskunganka @yasserf what are your thoughts?

yasserf:
NodeJS didn't do it  initially, but I just took a look through the current release history since the major releases became a thing and they stick to the books now ( also put labels next to commits which do minor and major changes which is quite cool for searching! ).

Then again react is still in 0.14 land, babel is in 6.5, angular in 1.5 and they have over a thousand releases between them. 

I'm for correct semantic versions in the long run. But since we are still in 0 land people should expect some potentially breaking changes as long as we make them very obvious in release notes.

PR #72, opened 2016-02-14
Fixes #27 anonymous record ready event

### LINKED ISSUE #27: Force disconnect/login a client that is connected

yasserf:


test-e2e/specs/anonymous-recordSpec.js
```diff
@@ -53,21 +53,21 @@ describe( 'anonymous record', function() {
...
-        }, 20);
+        });
+        anonymousRecord.setName( 'recordA' );
     });
     
     it( 'sets recordB', function(done) {
-        anonymousRecord.setName( 'recordB' );
-        setTimeout(function(){
-            expect( currentPet ).toBe( 'pug' );
+        anonymousRecord.once( 'ready', function(){
```
WolframHempel: Would be good to add a test that makes sure that this happens exactly once. I'll add one on a unit test level

WolframHempel:
@yasserf Looks good, I've added some unit tests as well. Could you have a last look and merge if happy

yasserf:
looks good

PR #73, opened 2016-02-15
Feature/login response data

yasserf:


src/message/connection.js
```diff
@@ -337,14 +339,16 @@ Connection.prototype._handleAuthResponse = function( message ) {
 		}
 		
 		if( this._authCallback ) {
-			this._authCallback( false, message.data[ 0 ], message.data[ 1 ] );
+			data = message.data[ 1 ] && messageParser.convertTyped( message.data[ 1 ], this._client );
```
yasserf: this will always be a string so loose comparison should be okay.. i think

PR #79, opened 2016-03-05
Bug/nested change doesnt trigger general subscribe

### LINKED ISSUE #78: record.subscribe without a path doesn't trigger for changes to nested objects
The problem is that `record._beginChange` uses `record.get` to retrieve `this._oldValue`, but `record.get` only uses `utils.shallowCopy`. Therefor `utils.deepEquals( this._oldValue, this._$data )` in `record._completeChange` returns true and the listeners are not notified.

I'll change it to deepCopy records. This is less performant, but also prevents hard to diagnose bugs such as unconsciously manipulating deepstream's internal values.

WolframHempel:
FIXES #78


src/utils/utils.js
```diff
@@ -1,88 +1,112 @@
...
 
+/**
+ * True if environment is node, false if it's a browser
+ * This seems somewhat inelegant, if anyone knows a better solution,
+ * let's change this (must identify browserify's pseudo node implementation though)
+ *
+ * @public
+ * @type {Boolean}
+ */
+exports.isNode = typeof process !== 'undefined' && process.toString() === '[object process]';
```
WolframHempel: Explanation: Didn't need to be a function that's evaluated everytime nextTick is called

src/utils/utils.js
```diff
@@ -1,88 +1,112 @@
...
+ *
+ * @public
+ * @returns {String} trimmedString
+ */
 exports.trim = function( inputString ) {
 	if( inputString.trim ) {
 		return inputString.trim();
 	} else {
-		return inputString.replace( trimRegExp, '' );
+		return inputString.replace( TRIM_REGULAR_EXPRESSION, '' );
```
yasserf: Is this quicker than .trim()?

src/utils/utils.js
```diff
@@ -1,88 +1,112 @@
...
-exports.deepEquals = function( objA, objB ) {
-	var isEqual = true, 
-		next;
-
-	if( objA === null || objB === null ) {
-		return objA === objB;
-	}
-
+/**
+ * Compares two objects for deep (recoursive) equality
```
yasserf: Recursive?

src/utils/utils.js
```diff
@@ -1,88 +1,112 @@
...
+ * @param   {Mixed} objB
+ *
+ * @public
+ * @returns {Boolean} isEqual
+ */
+exports.deepEquals= function( objA, objB ) {
 	if( typeof objA !== OBJECT || typeof objB !== OBJECT ) {
 		return objA === objB;
+	} else {
+		return JSON.stringify( objA ) === JSON.stringify( objB );
```
yasserf: I like, was thinking this would be a good approach too.

It doesnt copy over functions though, and I'm not sure about date objects ( on phone now so can't test )

src/utils/utils.js
WolframHempel: It's a polyfill for trim (see line 52/53) the only change is that I renamed the variable

src/utils/utils.js
WolframHempel: ah, is that not the British spelling (as in colour and honour?)

src/utils/utils.js
WolframHempel: You're right, tt doesn't, but it is only used to clone record data which is serializable anyways

src/utils/utils.js
yasserf: Ah my bad, didn't realise. Didn't know that trim was something new/ish

src/utils/utils.js
yasserf: agreed, but is it worth mentioning as well in docs? I'm happy eitherways ( also, this comment was meant to be for deepClone method )

Issue #80, opened 2016-03-05
Travis build seems to only run a small number of tests


WolframHempel:
The command it runs is correct, but it looks like it only runs about 20 tests (judging by the green dots) as opposed to the ~500 it should run.

![image](https://cloud.githubusercontent.com/assets/1503717/13546791/e8ffeaea-e2b8-11e5-9922-de74a3a40602.png)

yasserf:
We can add a reporter to make results look shiny

PR #89, opened 2016-03-26
Enhancement/upgrading jasmine 2.4

yasserf:


yasserf:
Things that 'regressed':

Jasmine .isEqual uses some... interesting comparison logic ( lifted from handlebars ) which compares objects constructors and returns false. Not entirely sure what the problem is but it only happened in two places so JSON.stringified for now after timebox

watch is no longer supported so removed it =(

Things that unregressed:

Finally found out why tests were randomly stopping. Turns out the util/cluster file would swallow the redis plugin init timeout and just exit everything, put a fix in which throws an error if anything critical happens on deepstream side, but could be done more elegantly.

We now use the official jasmine npm package which means we can keep up to date with jasmine versions now. Fixed the few spy changes but things were generally kept the same

We now have npm run-script reporter which gives us prettier tests and failures results, putting into build

Fixes #80

PR #90, opened 2016-03-28
Feature/record has and snapshot

yasserf:


yasserf:
Tests are failing since they depend on the deepstream server in e2e tests

Also missing unit tests for SingleRequestor to ensure architecture looks good, but that should be the only missing thing.

PR #98, opened 2016-04-24
Added deepcopy

### LINKED ISSUE #96: Deepstream not updating object properties
Deepstream will not push an update to subscribers of a record if that record is an object which has a property that changes. Offending code:
    var myObject = {hello:1};
    ds.record.getRecord('myRecord').set('mySubrecord',myObject);
    //mySubrecord on this client returns {hello:1}, other clients {hello:1}
    myObject.hello = 2;
    //mySubrecord on this client returns {hello:1}, other clients {hello:1}
    ds.record.getRecord('myRecord').set('mySubrecord',myObject);
    //mySubrecord on this client returns {hello:2}, other clients {hello:1}
This is specifically happening using the node.js client at 0.5.0, node.js version 5.11.0, and node.js server at 0.9.0.

drsirmrpresidentfathercharles:
Fixes #96

### DIFF
--- a/src/record/record.js
+++ b/src/record/record.js
@@ -110,14 +110,14 @@ Record.prototype.set = function( pathOrData, data ) {
 	this._version++;
 
 	if( arguments.length === 1 ) {
-		this._$data = pathOrData;
+		this._$data = utils.deepCopy(pathOrData);
 		this._connection.sendMsg( C.TOPIC.RECORD, C.ACTIONS.UPDATE, [
 			this.name,
 			this._version,
 			this._$data
 		]);
 	} else {
-		this._getPath( pathOrData ).setValue( data );
+		this._getPath( pathOrData ).setValue( utils.deepCopy(data) );
 		this._connection.sendMsg( C.TOPIC.RECORD, C.ACTIONS.PATCH, [
 			this.name,
 			this._version,
--- a/src/record/record.js
+++ b/src/record/record.js
@@ -110,14 +110,14 @@ Record.prototype.set = function( pathOrData, data ) {
 	this._version++;
 
 	if( arguments.length === 1 ) {
-		this._$data = utils.deepCopy(pathOrData);
+		this._$data = (typeof pathOrData == 'object') ? utils.deepCopy(pathOrData):pathOrData;
 		this._connection.sendMsg( C.TOPIC.RECORD, C.ACTIONS.UPDATE, [
 			this.name,
 			this._version,
 			this._$data
 		]);
 	} else {
-		this._getPath( pathOrData ).setValue( utils.deepCopy(data) );
+		this._getPath( pathOrData ).setValue((typeof data == 'object') ? utils.deepCopy(data):data);
 		this._connection.sendMsg( C.TOPIC.RECORD, C.ACTIONS.PATCH, [
 			this.name,
 			this._version,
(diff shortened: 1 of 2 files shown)


yasserf:
Thanks for the PR. Can you also please add a test that would have failed before the fix?

drsirmrpresidentfathercharles:
Hang on, my tests still failed. I'm new to Jasmine.

drsirmrpresidentfathercharles:
OK, so tests are off by one, which would lead me to believe it is due to the extra message I'm sending in the unit tests. I don't understand quite how those unit tests work, is there someone who does?

drsirmrpresidentfathercharles:
I looked at the unit tests and can't identify why they are failing. These tests passed with commit ae73134, which incorporated the deep copy of an object, but as soon as the new unit test was added correctly (deadd44), the tests for splitting up messages across packets started to fail by one buffered packet. From what I can identify, it appears that packed w4 is sending, packed w5 is sending, and then the rest of the packets are being buffered, so not only does the first packet send fail for the wrong packet, then sending all the buffered messages afterward fails because there is a missing packed from the beginning.

yasserf:
hey @drsirmrpresidentfathercharles, I'll take a look at this tomorrow and get back to you on that

yasserf:
Took a look into this and it is truly bizarre. Just adding a single test somehow breaks it. Tried to find out  the reason with no luck, but look again first thing tomorrow with my coffee.

yasserf:
@drsirmrpresidentfathercharles Hey, so while I wasn't able to find out the exact issue I found a solution.

So the problem itself is with timeouts. 

Tests that pass do this:
```
Sending EEVTw0 OPEN 1 5 1 5
Write: EEVTw0
Trace: Here I am! <-- The important part, triggered by process.nextTick
.next tick
Sending EEVTw1 OPEN 1 5 1 5
Write: EEVTw1
setting up tick
Sending EEVTw2 OPEN 1 5 2 5
Write: EEVTw2
.setting up tick
Sending EEVTw3 OPEN 1 5 3 5
Write: EEVTw3
setting up tick
Sending EEVTw4 OPEN 1 5 4 5
Write: EEVTw4
```

Tests that fail ( not no timeout):
```
next tick
Sending EEVTw0 OPEN 1 5 1 5
Write: EEVTw0
.setting up tick
Sending EEVTw1 OPEN 1 5 2 5
Write: EEVTw1
setting up tick
Sending EEVTw2 OPEN 1 5 3 5
Write: EEVTw2
.setting up tick
Sending EEVTw3 OPEN 1 5 4 5
Write: EEVTw3
setting up tick
Buffering EEVTw4 OPEN 1 5 5 5
```

The solution to this is to change [this line](https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/test-unit/unit/message/connectionSpec.js#L344) to `sendMessages( connection, 3, 8 );`

In summary, I have no idea why the process.nextTick callback is called. The solution provided is correct behavior. I'll keep an eye out for why moving on.

Otherwise have merged your PR across and made fix. thanks!

drsirmrpresidentfathercharles:
@yasserf Thank you. I just realized I was never able to run Grunt build (the unit tests would never pass), so the client js under dist/ will be incorrect until Grunt is run again.

PR #100, opened 2016-05-02
Feature/#1versionconflicts

yasserf:
Not finished, raising for Tests


src/constants/merge-strategies.js
```diff
@@ -0,0 +1,29 @@
+function merge( oldData, newData, callback ) {
+	var prop;
+	for( prop in newData ) {
+		if( typeof( newData[ prop ] ) === "object" ) {
+			if( !prop in oldData || typeof( oldData[ prop ] ) !== "object" ) {
+				// Safe override or addition
+				oldData[ prop ]= newData[ prop ]; 
+			} else {
+				// Recursive
+				oldData[ prop ] = this.merge( oldData[ prop ], newData[ prop ] );
```
WolframHempel: do we need the `this`?

src/constants/merge-strategies.js
```diff
@@ -0,0 +1,29 @@
+function merge( oldData, newData, callback ) {
```
WolframHempel: guess I'm missing something, but where's the no conflict bit?

src/record/record.js
```diff
@@ -264,7 +286,7 @@ Record.prototype._$onMessage = function( message ) {
 		this._applyUpdate( message, this._client );
 	}
 	else if( message.data[ 0 ] === C.EVENT.VERSION_EXISTS ) {
-		this._recoverRecord( message );
+		this._recoverRecord( message.data[ 2 ], JSON.parse( message.data[ 3 ] ), message );
```
WolframHempel: Could we used the `typed` mechanism for that

yasserf: Object being sent here isn't using typed since it is always JSON, same as read and update

PR #105, opened 2016-05-06
Feature/#1versionconflicts

yasserf:
This PR is now ready, but first required redirect logic to be merged in.

### DIFF
--- a/src/constants/merge-strategies.js
+++ b/src/constants/merge-strategies.js
@@ -1,12 +1,12 @@
 module.exports = {
 	/**
-	*	Choose the server state over the clients
+	*	Choose the server's state over the client's
 	**/
 	REMOTE_WINS: function( record, remoteValue, remoteVersion, callback ) {
 		callback( null, remoteValue );
 	},
 	/**
-	*	Choose the local state over the servers
+	*	Choose the local state over the server's
 	**/
 	LOCAL_WINS: function( record, remoteValue, remoteVersion, callback ) {
 		callback( null, record.get() );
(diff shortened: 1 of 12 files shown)


src/constants/merge-strategies.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+	/**
+	*	Choose the server state over the clients
```
WolframHempel: *grammar nazi* client's / server's (no need to actually change)

src/record/record.js
```diff
@@ -47,6 +52,27 @@ var Record = function( name, recordOptions, connection, options, client ) {
...
+ * completed or if an error occurs ( which leaves it in an inconsistent state until
+ * the next update merge attempt ).
+ *
+ * @param   {Function} mergeStrategy A Function that can resolve merge issues.
+ *
+ * @public
+ * @returns {void}
+ */
+Record.prototype.setMergeStrategy = function( mergeStrategy ) {
+	if( mergeStrategy instanceof Function ) {
```
WolframHempel: don't care too much personally, but there's been a few articles advising against checking instanceof primitives (e.g. when primitives are fiddled with in tests, when clients are distributed across browser windows etc... ) and rather use typeof... (similar thing with === undefined)

src/default-options.js
```diff
@@ -1,3 +1,5 @@
+var MERGE_STRATEGIES = require( './constants/merge-strategies' );
```
WolframHempel: needs to be exposed for users to access

WolframHempel:
Merging, but wondering if we should rather expose the constants on the instance as that's the only thing we pass to e.g. the react plugin

PR #117, opened 2016-05-17
Fix #75 reduce the bundle by ignoring the TCP connection class in bro…

### LINKED ISSUE #75: Reduce the dependencies within the client bundle
Webpack provides a sleeker version of browserify. It's recommended by react and the JavaScript world seems to be moving towards it.  It might help us reduce the size of the client files be bundling less unneeded dependencies.

http://webpack.github.io/

yasserf:
…wserify


WolframHempel:
Nice - can we make sure it still works with webpack?

yasserf:
Is webpack affected by the browserify config in grunt?

WolframHempel:
It's reading this bit from `package.json`

```
 "browser": {
    "net": false
  },
```

PR #122, opened 2016-05-22
Bug/#85 get list referencing

yasserf:


yasserf:
Build failing due to a dependency on permissions.json

Issue #138, opened 2016-05-31
subscribe is never called with the initial value on created but not set records?


ronag:
e.g.
```
const rec = ds.record.getRecord(id)
rec.subscribe(function (value) {
  console.log(value)
}, true)
```

Given that records are create on `getRecord` I would expect the above code to print `{}`...

Or is the record state after `getRecord` something in between exists and doesn't exists?

yasserf:
getRecord is async if the record isn't already loaded client side, which means you won't have the data from the server immediately afterwards when you do a subscribe( fn, true )

ronag:
@yasserf: Yes, but shouldn't it eventually print `{}`?

yasserf:
This is got to do with it just being async

```
const rec = ds.record.getRecord(id) <- async call to get data
rec.subscribe(function (value) {
  console.log(value)
}, true) <- immediate callback that gets triggered before record is created
```

```
const rec = ds.record.getRecord(id) <- async call to get data
ds.record.snapshot( function() { 
} ) <- also async, hences gets called with data
```

yasserf:
The subscribe method should print {} once the initial data comes back though. Is that not the case?

ronag:
>  Is that not the case?

No it's not...

ronag:
```
const rec = ds.record.getRecord(id) <- async call to get data
ds.record.snapshot( function() { 
} ) <- also async, hences gets called with data
```

It's not async, see: https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/record/record-handler.js#L138

It should be, at the least:

```
RecordHandler.prototype.snapshot = function( name, callback ) {
	if( this._records[ name ] ) {
		setImmediate(callback.bind(null, null, this._records[ name ].get() );
	} else {
		this._snapshotRegistry.request( name, callback );
	}
};
```

ronag:
The reason this is an issue for us is that we now have to use a rather silly implementation:

```
observe (recordName, path) {
    const record = ds.record.getRecord(recordName)
    return Observable
      .create(o => {
        record.whenReady(() => {
          o.next(record.get(path))
          o.complete()
        })
      })
      .concat(Observable
        .create(o => {
          const update = (value) => o.next(value)
          record.subscribe(path, update)
          return () => {
            record.unsubscribe(path, update)
            record.discard()
          }
        })
      )
  }
```

yasserf:
[...]

PR #142, opened 2016-05-31
record: Fixed bug with subscription missing initial value.

ronag:
Fixes https://github.com/deepstreamIO/deepstream.io-client-js/issues/138


ronag:
This is missing a corresponding unit test.

ronag:
This makes sense given the current behaviour of `set` not being visible to `get` unless `isReady`.

ronag:
@yasserf

PR #145, opened 2016-06-01
Global pattern matched merge strategies.

ronag:
DO NOT MERGE.

Work in progress/example of how we think the merge API can be improved.


yasserf:
This is quite interesting, we won't be putting it in the 1.0 release but if we have a few upvotes could plan it in sooner than later.

PR #187, opened 2016-07-26
Fixed broken unsubscribe.

ronag:
This fixes an issue where `unsubscribe(callback)` would unsubscribe **all** callbacks instead of just the one provided.


ronag:
This probably needs test as well

PR #195, opened 2016-07-27
Feature/#194 terminate unauthenticated connections

yasserf:
Requires (feature/#226-terminate-unauthenticated-connections) on server to be merged for the tests to pass


test-e2e/specs/recordSpec.js
```diff
@@ -36,12 +36,32 @@ describe( 'record', function() {
...
+		function immediateSubscribe( data ) {
+			clientARecord.unsubscribe( immediateSubscribe );
+			immediateSubscribeCalled = true;
+			expect( data ).toEqual( {} );
+		}
+		expect( clientARecord.isReady ).toBe( false );
+		clientARecord.subscribe( immediateSubscribe, true );
+		expect( immediateSubscribeCalled ).toBe( false );
+
+		var update = 1;
```
timaschew: IMHO it's easier to understand when this is a boolean with the name `self´ or sth like that

src/record/list.js
```diff
@@ -172,10 +172,13 @@ List.prototype.subscribe = function() {
 	}
 
 	//Make sure the callback is invoked with an empty array for new records
-	parameters.callback = function( callback ) {
+	var listCallback = function( callback ) {
 		callback( this.getEntries() );
 	}.bind( this, parameters.callback );
 
+	parameters.callback.wrappedCallback = listCallback;
+	parameters.callback = listCallback;
```
timaschew: This looks weird:
- setting a property of an object
- overwriting the whole object

Maybe create two variables for `_record. subscribe` and `_record.unsubscribe` instead of modifying the `parameters` variable.

test-e2e/specs/recordSpec.js
```diff
@@ -88,71 +108,77 @@ describe( 'record', function() {
...
+			function subscribePath( value ){
 				expect( value ).toBeUndefined();
 			};
-			clientA.record.getRecord( 'record1' ).subscribe( 'objectToDelete', subscribePath );
-			clientB.record.getRecord( 'record1' ).subscribe( 'objectToDelete', subscribePath );
+			clientARecord.subscribe( 'objectToDelete', subscribePath );
+			clientBRecord.subscribe( 'objectToDelete', subscribePath );
 
-			var subscribeObject = function( value ) {
+			function subscribeObject( value ) {
```
timaschew: `subscribeObject` is a suboptimal name

test-unit/unit/record/record-subscriptionSpec.js
```diff
@@ -38,16 +40,39 @@ describe( 'supscriptions to local record changes', function(){
...
+			record.unsubscribe( function() {} );
+
+			record.set( 'firstname', 'Ray' );
+			expect( generalCallback.calls.count() ).toEqual( 5 );
+			expect( anotherGeneralCallback ).toHaveBeenCalledWith({ firstname: 'Ray', lastname: 'Hempel' });
+			expect( anotherGeneralCallback.calls.count() ).toEqual( 6 );
+		});
+
+
+	anotherGeneralCallback
```
timaschew: What does this line do? 
copy-paste issue?

src/record/list.js
yasserf: Completely agree this is wierd. However we can't modify the record since it is a singleton.

test-e2e/specs/recordSpec.js
yasserf: Its a test that means subscribe to entire object as opposed to subscribe path

test-unit/unit/record/record-subscriptionSpec.js
yasserf: good catch!

test-e2e/specs/recordSpec.js
yasserf: this is using a number since there are three states
1 -> undefined
2 -> Wolfram
3+ -> don't care, but didn't unsubscribe

PR #206, opened 2016-08-09
record: Always call whenReady asynchronously. Fixes #204

### LINKED ISSUE #204: discard/delete data race when quickly re-subscribing
Because in most cases it is not. Currently I have an application where often a record is discarded before it receives the initial value and I'm spammed with these errors.

ronag:
I'm not sure why this is required. But without this weird stuff happens. See #204.


ronag:
Doesn't fix the issue.

PR #209, opened 2016-08-09
Fix record discard/delete & subscribe race

ronag:
This is my try at fixing the discard/delete and subscribe data-races. https://github.com/deepstreamIO/deepstream.io-client-js/issues/204

The basic idea is that the record can revive itself AFTER successful discard/delete, using the pending queue for the time between re-subscription and revival.

This ensures that there CAN BE ONLY ONE active record instance for the client avoiding the:

```
UNSOLICITED_MESSAGE
UNSOLICITED_MESSAGE
ACK_TIMEOUT
RESPONSE_TIMEOUT
MULTIPLE_SUBSCRIBTIONS
```

error cycle

### DIFF
--- a/src/record/record-handler.js
+++ b/src/record/record-handler.js
@@ -42,9 +42,7 @@ RecordHandler.prototype.getRecord = function( name, recordOptions ) {
 	if( !this._records[ name ] ) {
 		this._records[ name ] = new Record( name, recordOptions || {}, this._connection, this._options, this._client );
 		this._records[ name ].on( 'error', this._onRecordError.bind( this, name ) );
-		this._records[ name ].on( 'destroyPending', this._onDestroyPending.bind( this, name ) );
-		this._records[ name ].on( 'delete', this._removeRecord.bind( this, name ) );
-		this._records[ name ].on( 'discard', this._removeRecord.bind( this, name ) );
+		this._records[ name ].on( 'destroy', this._removeRecord.bind( this, name ) );
 	}
 
 	this._records[ name ].usages++;
@@ -185,26 +183,6 @@ RecordHandler.prototype._$handle = function( message ) {
 	if( message.action === C.ACTIONS.ACK || message.action === C.ACTIONS.ERROR ) {
 		name = message.data[ 1 ];
 
-		/*
-		 * The following prevents errors that occur when a record is discarded or deleted and
-		 * recreated before the discard / delete ack message is received.
-		 *
-		 * A (presumably unsolvable) problem remains when a client deletes a record in the exact moment
-		 * between another clients creation and read message for the same record
-		 */
-		if( message.data[ 0 ] === C.ACTIONS.DELETE ||
-			  message.data[ 0 ] === C.ACTIONS.UNSUBSCRIBE ||
-			 ( message.data[ 0 ] === C.EVENT.MESSAGE_DENIED && message.data[ 2 ] === C.ACTIONS.DELETE  )
-			) {
-			this._destroyEventEmitter.emit( 'destroy_ack_' + name, message );
-
-			if( message.data[ 0 ] === C.ACTIONS.DELETE && this._records[ name ] ) {
-				this._records[ name ]._$onMessage( message );
-			}
-
-			return;
-		}
-
 		if( message.data[ 0 ] === C.ACTIONS.SNAPSHOT ) {
 			message.processedError = true;
 			this._snapshotRegistry.recieve( name, message.data[ 2 ] );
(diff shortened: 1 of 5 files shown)


src/record/record-handler.js
```diff
@@ -185,26 +183,6 @@ RecordHandler.prototype._$handle = function( message ) {
...
-			) {
-			this._destroyEventEmitter.emit( 'destroy_ack_' + name, message );
-
-			if( message.data[ 0 ] === C.ACTIONS.DELETE && this._records[ name ] ) {
-				this._records[ name ]._$onMessage( message );
-			}
-
-			return;
-		}
-
```
ronag: Parts of this is probably still needed. Unclear.

src/record/record.js
```diff
@@ -237,9 +240,10 @@ Record.prototype.discard = function() {
 	this.whenReady( function() {
 		this.usages--;
 		if( this.usages <= 0 ) {
-				this.emit( 'destroyPending' );
-				this._discardTimeout = setTimeout( this._onTimeout.bind( this, C.EVENT.ACK_TIMEOUT ), this._options.subscriptionTimeout );
-				this._connection.sendMsg( C.TOPIC.RECORD, C.ACTIONS.UNSUBSCRIBE, [ this.name ] );
+			this.isReady = false;
+			this._discardTimeout = setTimeout( this._onTimeout.bind( this, C.EVENT.ACK_TIMEOUT ), this._options.subscriptionTimeout );
+			this._connection.sendMsg( C.TOPIC.RECORD, C.ACTIONS.UNSUBSCRIBE, [ this.name ] );
+	  	this._resubscribeNotifier.destroy();
```
ronag: Not part of the original implementation. But I believe the resubscribeNotifier should be destroyed as soon as the `UNSUBSCRIBE`has been sent.

src/record/record.js
```diff
@@ -628,14 +633,18 @@ Record.prototype._onTimeout = function( timeoutType ) {
...
- 	this._clearTimeouts();
- 	this._eventEmitter.off();
- 	this._resubscribeNotifier.destroy();
- 	this.isDestroyed = true;
- 	this.isReady = false;
- 	this._client = null;
-	this._eventEmitter = null;
-	this._connection = null;
+	if (this.usages > 0) {
+		this._init()
```
ronag: This is the magic. We allow a record to be revive AFTER it's life cycle has finished.

src/record/record.js
```diff
@@ -255,9 +259,10 @@ Record.prototype.delete = function() {
 		return;
 	}
 	this.whenReady( function() {
-		this.emit( 'destroyPending' );
+		this.isReady = false;
```
ronag: Setting isReady to false allows the record to be revived with pending operations.

src/record/record.js
```diff
@@ -237,9 +240,10 @@ Record.prototype.discard = function() {
 	this.whenReady( function() {
 		this.usages--;
 		if( this.usages <= 0 ) {
-				this.emit( 'destroyPending' );
-				this._discardTimeout = setTimeout( this._onTimeout.bind( this, C.EVENT.ACK_TIMEOUT ), this._options.subscriptionTimeout );
-				this._connection.sendMsg( C.TOPIC.RECORD, C.ACTIONS.UNSUBSCRIBE, [ this.name ] );
+			this.isReady = false;
```
ronag: Setting isReady to false allows the record to be revived with pending operations.

ronag:
travis fails since the test don't allow re-using/reviving record instances.

ronag:
This also resolve the issue of spamming subscribe/unsubscribe as there as a natural throttling.

src/record/record.js
```diff
@@ -628,14 +628,19 @@ Record.prototype._onTimeout = function( timeoutType ) {
...
- 	this.isDestroyed = true;
- 	this.isReady = false;
- 	this._client = null;
-	this._eventEmitter = null;
-	this._connection = null;
+	if (this.usages > 0) {
+		this._sendRead();
+	}
+	else {
+		this._clearTimeouts();
```
ronag: Clear timeout beredda to be moved to top and always be invoked.

ronag:
@yasserf: tests pass

ronag:
@yasserf: This PR has a race when disconnecting the client. I'm currently in a state where I have no working solution to this issue.

ronag:
I'm closing this as it's broken.

ronag:
Fixed issue.

src/utils/resubscribe-notifier.js
```diff
@@ -40,14 +40,14 @@ ResubscribeNotifier.prototype.destroy = function() {
...
-		
-	if( state === C.CONNECTION_STATE.RECONNECTING && this._isReconnecting === false ) {
-		this._isReconnecting = true;
-	}
+
 	if( state === C.CONNECTION_STATE.OPEN && this._isReconnecting === true ) {
 		this._isReconnecting = false;
 		this._resubscribe();
 	}
+	else if ( this._isReconnecting === false ) {
```
ronag: We always want to reconnect when going from non OPEN to OPEN.

ronag:
#232

src/record/record.js
```diff
@@ -42,11 +42,7 @@ var Record = function( name, recordOptions, connection, options, client ) {
 	if( options.mergeStrategy ) {
 		this.setMergeStrategy( options.mergeStrategy );
 	}
-
-	this._resubscribeNotifier = new ResubscribeNotifier( this._client, this._sendRead.bind( this ) );
-	this._readAckTimeout = setTimeout( this._onTimeout.bind( this, C.EVENT.ACK_TIMEOUT ), this._options.recordReadAckTimeout );
-	this._readTimeout = setTimeout( this._onTimeout.bind( this, C.EVENT.RESPONSE_TIMEOUT ), this._options.recordReadTimeout );
-	this._sendRead();
+	this._resubscribeNotifier = new ResubscribeNotifier( this._client, this._sendRead.bind( this ), this._clearTimeouts.bind( this ), true );
```
ronag: Simply doing a clearTimeouts on disconnect is not good enough...

ronag:
I don't think this PR is good enough. But the basic ideas are there. There are some edge cases that I don't think are properly handled.

yasserf:
Yeah, we need to look into the discard/delete async case in more detail

ronag:
This needs to be solved in another way. We've resolved it locally in our fork in a different way.

Issue #213, opened 2016-08-18
path listeners are compared by reference instead of deeply


ronag:
i.e.

```javascript
	for( path in this._oldPathValues ) {
		currentValue = this._getPath( path ).getValue();

		if( currentValue !== this._oldPathValues[ path ] ) {
			this._eventEmitter.emit( path, currentValue );
		}
	}
```

Should be:

```javascript
	for( path in this._oldPathValues ) {
		currentValue = this._getPath( path ).getValue();

		if( !utils.deepEquals(currentValue, this._oldPathValues[ path ] ) ) {
			this._eventEmitter.emit( path, currentValue );
		}
	}
```

ronag:
Fixed in #214

ronag:
@yasserf: bug

ronag:
Fixed

PR #200, opened 2016-08-04
requestIdleCallback

ronag:
This makes deepstream heavy web applications much more responsive.

PR #214, opened 2016-08-18
Optimise & Refactor Records

### LINKED ISSUE #172: Feature Request - Don't Perform Deep Copies
Right now deepstream will perform deep copies of all records when requesting data. We would prefer if it just called `Object.freeze` or some similar solution.

ronag:
I know that currently this is not something that will get into deepstream but I would at least ask you to consider it, and possibly give some feedback, since we need it.

Our application is very record heavy and we have some problems with keeping stable 60 fps. We've already improved on this issue with #200 but we still have some performance issues.

A part form general refactoring and simplification:

Optimizations:

- Optional deep copies. Assume immutability.
- Cache path tokenisation.
- Reduced allocations.
- Reduced copies.

Fixes #172, #213

We could make even more use of `Object.create( null )` over `{}` for jsonPath. However, jasmine test for types in `isEqual` which would make some tests fail.

### DIFF
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -12,10 +12,9 @@ var cache = Object.create( null );
  * @returns {Mixed}
  */
 module.exports.get = function ( data, path, deepCopy ) {
-	var tokens = tokenize( path ),
-		i;
+	var tokens = tokenize( path );
 
-	for( i = 0; i < tokens.length; i++ ) {
+	for( var i = 0; i < tokens.length; i++ ) {
 		if( data[ tokens[ i ] ] !== undefined ) {
 			data = data[ tokens[ i ] ];
 		} else {
@@ -36,20 +35,19 @@ module.exports.get = function ( data, path, deepCopy ) {
  * @returns old or new state
  */
 module.exports.set = function( data, path, value, deepCopy ) {
-	var i,
-		tokens = tokenize( path );
+	var tokens = tokenize( path );
 
 	if ( deepCopy !== false ) {
 		value = utils.deepCopy( value );
 	}
 
 	if ( tokens.length > 0 ) {
-		data = utils.shallowCopy( data );
-
-		var node = data;
-
-		for( i = 0; i < tokens.length - 1; i++ ) {
-			if( node[ tokens[ i ] ] !== undefined ) {
+		var node = data = utils.shallowCopy( data );
+		for( var i = 0; i < tokens.length; i++ ) {
+			if ( i === tokens.length - 1) {
+				node[ tokens[ i ] ] = value;
+			}
+			else if( node[ tokens[ i ] ] !== undefined ) {
 				node = node[ tokens[ i ] ] = utils.shallowCopy( node[ tokens[ i ] ] );
 			}
 			else if( tokens[ i + 1 ] && !isNaN( tokens[ i + 1 ] ) ){
@@ -59,8 +57,6 @@ module.exports.set = function( data, path, value, deepCopy ) {
 				node = node[ tokens[ i ] ] = {};
 			}
 		}
-
-		node[ tokens[ i ] ] = value;
 	}
 	else {
 		data = value;
@@ -76,22 +72,15 @@ module.exports.set = function( data, path, value, deepCopy ) {
  * @returns Array of tokens
  */
 function tokenize( path ) {
-	path = path ? path.toString() : undefined
-
-	var tokens = cache[ path ];
-
-	if ( tokens ) {
-		return tokens;
+	if ( cache[ path ] ) {
+		return cache[ path ];
 	}
 
-	tokens = [];
-
-	var parts = path ? path.split( SPLIT_REG_EXP ) : [],
+	var parts = path !== undefined ? path.toString().split( SPLIT_REG_EXP ) : [],
 		tokens = [],
-		part,
-		i;
+		part;
 
-	for( i = 0; i < parts.length; i++ ) {
+	for( var i = 0; i < parts.length; i++ ) {
 		part = utils.trim( parts[ i ] );
 
 		if( part.length === 0 ) {
@@ -111,7 +100,5 @@ function tokenize( path ) {
 		tokens.push( part );
 	}
 
-	cache[ path ] = tokens;
-
-	return tokens;
+	return cache[ path ] = tokens;
 };
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -107,7 +107,7 @@ function tokenize( path ) {
 		return cache[ path ];
 	}
 
-	var parts = path !== undefined ? String( path ).match(PARTS_REG_EXP) : [];
+	var parts = String(path) !== 'undefined' ? String( path ).match(PARTS_REG_EXP) : [];
 
 	if ( !parts ) {
 		throw new Error('invalid path ' + path)
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -41,25 +41,24 @@ module.exports.set = function( data, path, value, deepCopy ) {
 		value = utils.deepCopy( value );
 	}
 
-	if ( tokens.length > 0 ) {
-		var node = data = utils.shallowCopy( data );
-		for( var i = 0; i < tokens.length; i++ ) {
-			if ( i === tokens.length - 1) {
-				node[ tokens[ i ] ] = value;
-			}
-			else if( node[ tokens[ i ] ] !== undefined ) {
-				node = node[ tokens[ i ] ] = utils.shallowCopy( node[ tokens[ i ] ] );
-			}
-			else if( tokens[ i + 1 ] && !isNaN( tokens[ i + 1 ] ) ){
-				node = node[ tokens[ i ] ] = [];
-			}
-			else {
-				node = node[ tokens[ i ] ] = {};
-			}
-		}
+	if ( tokens.length === 0 ) {
+		return value;
 	}
-	else {
-		data = value;
+
+	var node = data = utils.shallowCopy( data );
+	for( var i = 0; i < tokens.length; i++ ) {
+		if ( i === tokens.length - 1) {
+			node[ tokens[ i ] ] = value;
+		}
+		else if( node[ tokens[ i ] ] !== undefined ) {
+			node = node[ tokens[ i ] ] = utils.shallowCopy( node[ tokens[ i ] ] );
+		}
+		else if( tokens[ i + 1 ] && !isNaN( tokens[ i + 1 ] ) ){
+			node = node[ tokens[ i ] ] = [];
+		}
+		else {
+			node = node[ tokens[ i ] ] = Object.create( null );
+		}
 	}
 
 	return data;
@@ -76,12 +75,12 @@ function tokenize( path ) {
 		return cache[ path ];
 	}
 
-	var parts = path !== undefined ? path.toString().split( SPLIT_REG_EXP ) : [],
-		tokens = [],
-		part;
+	var parts = path !== undefined ? path.toString().split( SPLIT_REG_EXP ) : [];
+
+	var tokens = [];
 
 	for( var i = 0; i < parts.length; i++ ) {
-		part = utils.trim( parts[ i ] );
+		var part = utils.trim( parts[ i ] );
 
 		if( part.length === 0 ) {
 			continue;
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -13,12 +13,8 @@ var cache = Object.create( null );
 module.exports.get = function ( data, path, deepCopy ) {
 	var tokens = tokenize( path );
 
-	for( var i = 0; i < tokens.length; i++ ) {
-		if( data[ tokens[ i ] ] !== undefined ) {
-			data = data[ tokens[ i ] ];
-		} else {
-			return undefined;
-		}
+	for( var i = 0; data && i < tokens.length; i++ ) {
+		data = data[ tokens[ i ] ];
 	}
 
 	return deepCopy !== false ? utils.deepCopy( data ) : data;
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -21,7 +21,7 @@ module.exports.get = function ( data, path, deepCopy ) {
 			data = data[ tokens[ i ] ];
 		}
 		else {
-			throw new Error( 'invalid data' );
+			throw new Error( 'invalid data or path' );
 		}
 	}
 
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -13,8 +13,16 @@ var cache = Object.create( null );
 module.exports.get = function ( data, path, deepCopy ) {
 	var tokens = tokenize( path );
 
-	for( var i = 0; data && i < tokens.length; i++ ) {
-		data = data[ tokens[ i ] ];
+	for( var i = 0; i < tokens.length; i++ ) {
+		if ( data === undefined ) {
+			return undefined;
+		}
+		else if ( typeof data === 'object' ) {
+			data = data[ tokens[ i ] ];
+		}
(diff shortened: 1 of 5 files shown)


yasserf:
Hey Ronag,

Interesting PR. I like the refactor for the path, I done the logic similar for the java client.

Just had an interesting conversation about `object.create( null )` which means we will officially drop IE8 support ( unless an external polyfill is provided ). That's fine though so its good.

Regarding deepCopy, we have alot of less experienced developers assume that a copy is provided back. If you allow `deepCopy` ( default true ) to be an option you pass into deepstreamClient on load that will be acceptable, but we can't change it without it.

> No deep equals. Assume reference equality and that set is only called when actually changed.

That's interesting, since yes it makes it faster but it also means listeners would get notified with same data constantly if your setting equal data. Adding another flag would also be beneficial, ie `deepCompare`

If you could also give some before and after metrics to updates that would interesting.

Cheers!

ronag:
> That's interesting, since yes it makes it faster but it also means listeners would get notified with same data constantly if your setting equal data.

In our case we have to perform that check anyway upstream in our application. So we basically end up with a lot of unnecessary checks.

ronag:
@yasserf: Take a look now. I've made the behaviour configurable.

ronag:
This ended up being a major refactoring & optimisation.

ronag:
@yasserf: tests pass

src/record/json-path.js
```diff
@@ -49,36 +33,66 @@ JsonPath.prototype.getValue = function() {
...
+	if (deepCompare !== false ? utils.deepEquals( oldValue, value ) : oldValue === value) {
+		return data;
+	}
+
+	if ( deepCopy !== false ) {
+		value = utils.deepCopy( value );
+	}
+
+	if ( tokens.length > 0 ) {
+		data = utils.deepCopy( data );
```
ronag: Should perform shallow copies if `deepCopy === false`. Future optimisation.

src/record/record.js
```diff
@@ -427,9 +419,9 @@ Record.prototype._onRead = function( message ) {
  */
 Record.prototype._setReady = function() {
 	this.isReady = true;
-	for( var i = 0; i < this._queuedMethodCalls.length; i++ ) {
-		this[ this._queuedMethodCalls[ i ].method ].apply( this, this._queuedMethodCalls[ i ].args );
-	}
+	this._queuedMethodCalls.forEach( function(call) {
```
yasserf: We don't use for each functions on hot paths. They unfortunately are not optimised as well as for loops, both from a memory and CPU aspect.

ronag: I'll change it.

yasserf:
This looks really interesting/ good. Given the amount of core changes I'll have to put some time aside to merge this in, but will aim for it to be in 1.1 coming soon!

src/record/json-path.js
```diff
@@ -49,58 +32,74 @@ JsonPath.prototype.getValue = function() {
...
+	}
 
-	for( i = 0; i < parts.length; i++ ) {
-		part = utils.trim( parts[ i ] );
+	var parts = path !== undefined ? path.toString().split( SPLIT_REG_EXP ) : [];
+
+	var tokens = [];
+
+	for( var i = 0; i < parts.length; i++ ) {
+		var part = utils.trim( parts[ i ] );
```
ronag: Do we really need trim?

src/record/json-path.js
```diff
@@ -49,58 +32,74 @@ JsonPath.prototype.getValue = function() {
...
+ * @returns Array of tokens
  */
-JsonPath.prototype._tokenize = function() {
-	var parts = this._path.split( SPLIT_REG_EXP ),
-		part,
-		i;
+function tokenize( path ) {
+	if ( cache[ path ] ) {
+		return cache[ path ];
+	}
```
ronag: @yasserf: Since the tokens are cached. Maybe we can afford some validation here and throw if the path is invalid?

src/record/json-path.js
```diff
@@ -49,58 +32,74 @@ JsonPath.prototype.getValue = function() {
...
 
-	for( i = 0; i < parts.length; i++ ) {
```
ronag: @yasserf: Since the tokens are cached. Maybe we can afford some validation here and throw if the path is invalid? Instead of split we could use a match.

src/record/json-path.js
yasserf: I think that's used for 'a[ 0 ]. ' paths, which should be supported

src/record/json-path.js
yasserf: Interesting point! That could work. Does it currently just throw an error?

ronag: @yasserf: I would say the current behaviour for invalid paths is undefined.

src/record/json-path.js
ronag: Then I would suggest a `paths.replace(/\s/g, '')` instead of doing it for each part.

src/record/json-path.js
```diff
@@ -49,58 +32,74 @@ JsonPath.prototype.getValue = function() {
...
+			tokens.push( parseInt( parts[ i ], 10 ) );
 			continue;
 		}
 
-		if( part === ASTERISK ) {
-			this._tokens.push( true );
+		if( parts[ i ] === ASTERISK ) {
+			tokens.push( true );
 			continue;
 		}
```
ronag: @yasserf: What does this do?

src/record/json-path.js
```diff
@@ -49,58 +32,74 @@ JsonPath.prototype.getValue = function() {
...
+		if( parts[ i ].length === 0 ) {
 			continue;
 		}
 
-		if( !isNaN( part ) ) {
-			this._tokens.push( parseInt( part, 10 ) );
+		if( !isNaN( parts[ i ] ) ) {
+			tokens.push( parseInt( parts[ i ], 10 ) );
 			continue;
 		}
```
ronag: I don't think parsing to int is required. We could just skip this?

src/record/record.js
```diff
@@ -125,34 +113,31 @@ Record.prototype.set = function( pathOrData, data ) {
...
 	}
 
-	if( arguments.length === 2 && utils.deepEquals( this._getPath( pathOrData ).getValue(), data ) ) {
-		return this;
-	}
-	else if( arguments.length === 1 && utils.deepEquals( this._$data, pathOrData ) ) {
+	var path = arguments.length === 1 ? undefined : pathOrData;
+	data = path ? data : pathOrData;
+
+	if ( utils.deepEquals( data, jsonPath.get( this._$data, path ) ) ) {
```
ronag: This check can become a duplicate if there is a listener for ALL_EVENT. Not sure how to resolve that in a clean fashion.

src/record/record.js
```diff
@@ -470,81 +437,31 @@ Record.prototype._setReady = function() {
...
-		currentValue = this._getPath( path ).getValue();
+	for ( var i = 0; i < paths.length; i++ ) {
+		var newValue = jsonPath.get( newData, paths[ i ], false );
+		var oldValue = jsonPath.get( oldData, paths[ i ], false );
 
-		if( currentValue !== this._oldPathValues[ path ] ) {
-			this._eventEmitter.emit( path, currentValue );
+		if( newValue !== oldValue && !utils.deepEquals( newValue, oldValue ) ) {
+			this._eventEmitter.emit( paths[ i ], this.get( paths[ i ] ) );
 		}
```
ronag: This could be further optimised by check the deepest paths first and going upwards, which would allow skipping checks for shallower paths if a descendant changed.

src/record/record.js
```diff
@@ -345,10 +324,13 @@ Record.prototype._recoverRecord = function( remoteVersion, remoteData, message )
  * @private
  * @returns {void}
  */
-Record.prototype._onRecordRecovered = function( remoteVersion, error, data ) {
+Record.prototype._onRecordRecovered = function( remoteVersion, remoteData, error, data ) {
 	if( !error ) {
 		this.version = remoteVersion;
-		this.set( data );
+		if ( data !== this._$data && !utils.deepEquals( data, this._$data ) ) {
```
ronag: @yasserf: I would argue that this should be:

```javascript
if ( data !== remoteData && !utils.deepEquals( data, remoteData ) ) {
```

However, the tests disagree with me. The current behaviour is strange to me, why would we send to the server when the resolved data is the same as the remoteData, but we don't send when the resolved data is the same as the local one?

Isn't this a bug?

src/record/record.js
```diff
@@ -125,34 +113,15 @@ Record.prototype.set = function( pathOrData, data ) {
...
-	} else {
-		this._getPath( pathOrData ).setValue( ( typeof data == 'object' ) ? utils.deepCopy( data ): data );
-		this._connection.sendMsg( C.TOPIC.RECORD, C.ACTIONS.PATCH, [
-			this.name,
-			this.version,
-			pathOrData,
-			messageBuilder.typed( data )
-		]);
+	if ( utils.deepEquals( data, jsonPath.get( this._$data, path ) ) ) {
+		return this;
```
ronag: This check ends up as a duplicate if there is a listener on that path. Not sure how to resolve that...

src/record/record.js
ronag: Also, what is the proper behaviour for `_onRecordRecovered` and `discard`.

Currently you could `discard` while the record is still being merged...

ronag: There are quite a bit of asynchronous issues here... I think this could be better written in terms of a state machine (ala Redux)...

ronag: On one more note. There currently is no timeout for the merge strategy callback...

yasserf: I'm confused. Code on master does not have a check..

https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/record/record.js#L351

ronag: @yasserf: `set` does the check...

ronag: Basically I don't think the "ours" merge strategy works with the current master...

yasserf: Gotcha, calling set traps the local incorrect state.

On a side note, this PR is getting huge. I'm not against most of the changes but be great to not do any actual functionality changes as it's going to be hard enough to test!

src/record/json-path.js
yasserf: This is legacy code, can be removed

src/record/record.js
ronag: @yasserf: Understood. I think all current behaviour is maintained and the tests pass. I've already made a separate issue for this problem.

yasserf: Yep, looking through all changes and noted the issue. This PR looks super promising and will be looking to merge it sometime this week hopefully!

src/record/json-path.js
```diff
@@ -49,58 +27,58 @@ JsonPath.prototype.getValue = function() {
...
+		else if( tokens[ i + 1 ] && !isNaN( tokens[ i + 1 ] ) ){
+			node = node[ tokens[ i ] ] = [];
 		}
 		else {
-			node = node[ this._tokens[ i ] ] = {};
+			node = node[ tokens[ i ] ] = Object.create( null );
 		}
 	}
 
-	node[ this._tokens[ i ] ] = value;
```
ronag: Using the immutable form of set should make it quite easy and efficient to later calculate deltas.

ronag:
Not sure why this is failing all of a sudden. All test pass fine on my local copy.

```
Critical error occured on deepstream PLUGIN_ERROR Error from messageConnector plugin: REDIS error:Received message for unknown topic CL
```

ronag:
Note that this still has the "bad" behaviour (see #224) from the previous implementation.

src/record/record.js
```diff
@@ -345,10 +327,19 @@ Record.prototype._recoverRecord = function( remoteVersion, remoteData, message )
...
+
+		var oldValue = this._$data;
+		var newValue = jsonPath.set( oldValue, undefined, data, false );
+
+		if ( oldValue === newValue ) {
+			return;
+		}
+
+		this._sendUpdate( undefined, data );
+		this._applyChange( newValue );
```
ronag: To fix #224 this should be:
```javascript
var oldLocalValue = this._$data;
var newLocalValue = jsonPath.set( oldValue, undefined, data, false );

if ( newLocalValue !== oldLocalValue ) {
  this._applyChange( newLocalValue )
}

var oldRemoteValue = remoteData;
var newRemoteValue = jsonPath.set( oldRemoteValue, undefined, data, false );
if ( newRemoteValue !== oldRemoteValue ) {
  this._sendUpdate( undefined, newRemoteValue )
}
```

yasserf:
I'm very keen on getting this PR merged and used as a baseline for the other async cases you mentioned. We have been doing some performance tests on deepstream and listening this entire week to ensure things work under heavy load and because of that we haven't been able to look into it yet.

I will most likely merge this in once I guarantee tests are passing, and release it under a 1.1RC along side listening while we continue doing stress testing over the next couple of weeks. I apologise for the time mixup / delay, given the amount of new features in 1.1 we want to thoroughly test it with all our internal use cases before we release it into the wild!

src/record/record.js
```diff
@@ -183,16 +155,12 @@ Record.prototype.subscribe = function( path, callback, triggerNow ) {
...
-			if( args.path ) {
-				args.callback( this._getPath( args.path ).getValue() );
-			} else {
-				args.callback( this.get() );
-			}
-		}.bind(this));
+		this.whenReady( function () {
+			this._eventEmitter.on( args.path, args.callback );
+			args.callback( this.get( args.path ) );
+		}.bind(this) );
```
ronag: I think in order to resolve #229 I think this should be:
```javascript
this._eventEmitter.on( args.path, args.callback );
if ( this.isReady ) {
  args.callback( this.get( args.path ) );
}
```

src/record/json-path.js
```diff
@@ -49,58 +33,87 @@ JsonPath.prototype.getValue = function() {
...
+	return result;
 };
 
+function patch( oldValue, newValue, deepCopy ) {
+	var i;
+
+	if ( utils.deepEquals( oldValue, newValue ) ) {
```
yasserf: I'm a little confused, won't this run recursively every time and outweigh the benefit of having the deepCopy flag? 

I can see it being an advantage when the flag is false since it will do literal object comparison, but when the flag is true it will have to do this on every object..


```
return deepCopy !== false ? utils.deepCopy( newValue ) : newValue;
```

src/record/record.js
yasserf: Makes sense!

src/record/json-path.js
ronag: @yasserf: Not sure what you are worried about? The extra comparison?

yasserf: The extra comparisonS.

deepEqual does a stringify twice. Since this is a recursive function it will do so on each individual item within the path.

If you compare the disadvantage of that with anyone who uses the `deepCopy` as default: `true`
to the advantage of when it is `false` I'm wondering how much more of a hit we are adding to the default usecase as a result.

ronag: The point here is also to maintain reference equality (even if deepCopy is true).

ronag: Essentially, this allows that any e.g. React component that is backed by a deepstream record can use `shallowCompare`.

ronag: @yasserf: You mean the overlap of deepEquals. Yes, that is a future improvement I documented somewhere.

yasserf: > The point here is also to maintain reference equality (even if deepCopy is true). patch should maintain references if the new value is equal.

Gotcha.

ronag: @yasserf: It would only be a problem in practice for large deeply nested records...

yasserf:
Done a rebase onto master and added missing documentation.

Merging it via #239

PR #220, opened 2016-08-22
Record/patch

ronag:
This continues from #214. By patching the existing `_$data` only on diffs we can always use reference equality instead of `deepEquals`. This makes things like the subscribe callback check, delta calculation and react (pure) components backed by records much faster.

The  `jsonPath.patch` can be further optimised later as an iterative implementation, which would be faster (for deeply nested records) than the recursive one in this PR. Further optimisation would also be to avoid doing overlapping equality checks.

### DIFF
--- a/src/default-options.js
+++ b/src/default-options.js
@@ -194,7 +194,5 @@ module.exports = {
 	 */
 	mergeStrategy: MERGE_STRATEGIES.REMOTE_WINS,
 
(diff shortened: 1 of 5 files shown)


ronag:
@yasserf: I put this in a separate PR in order to not make the other one bigger. Not sure if you want this in #214 or separate.

ronag:
rebased into #214

PR #221, opened 2016-08-23
Allow set while not isReady.

ronag:
This is mostly to start a discussion.

By using merge strategies we no longer need to defer sets while `!isReady`. Instead allow sets while `!isReady` but don't send update to the server. Later when the initial value arrives from the server we can use the merge strategy to consolidate the diff.

### DIFF
--- a/src/record/record.js
+++ b/src/record/record.js
@@ -36,7 +36,6 @@ var Record = function( name, recordOptions, connection, options, client ) {
 	this._oldValue = null;
 	this._oldPathValues = null;
 	this._eventEmitter = new EventEmitter();
-	this._queuedMethodCalls = [];
 
 	this._mergeStrategy = null;
 	if( options.mergeStrategy ) {
@@ -120,11 +119,6 @@ Record.prototype.set = function( pathOrData, data ) {
 		return this;
 	}
 
-	if( !this.isReady ) {
-		this._queuedMethodCalls.push({ method: 'set', args: arguments });
-		return this;
-	}
-
 	if( arguments.length === 2 && utils.deepEquals( this._getPath( pathOrData ).getValue(), data ) ) {
 		return this;
 	}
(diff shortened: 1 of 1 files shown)


ronag:
This will later also enable offline mode...

ronag:
This is nice for a few reasons:

- Simplified code by using a generic update path (i.e. _onRead is not required).
- Allows sets directly.

Some problems:

- Needs a smarter merge strategy as the default ones are not enough (maybe MERGE_LOCAL?).
- Cannot be made to work due to #224.

ronag:
Conceptually I don't think this is a very difficult thing. However, in practice and with tests I don't think I will have the time to finish this up by myself.

yasserf:
I commented on this a while ago but I think my mobile connection dropped and the comment vanished ( the irony ).

I pretty much wrote the same thing you as your last comment. Starting to edit something that isn't ready just leads to a whole other set of issues. Probably not for users who are aware of the internals, but if its the first time you use the API ( or the 100th if you don't care about the async nature entirely ) then it would be weird getting notified of the local state of the object that as opposed to the server state + the changes made. I can see this resulting in more issues than advantages.

I really think this story should be done when we investigate offline first ( where you would load the last record from a client side cache first ), although that is still a significant time away.

ronag:
> then it would be weird getting notified of the local state of the object that as opposed to the server state + the changes made. 

I would have to disagree. Fundamentally doing it like I am proposing should conceptually be exactly the same as not doing it, i.e. the user can receive new state from the server whenever, whether it is the "first" (if there is such a thing in a distributed system) or a later state is fundamentally the same.

Engineering wise I think this resolves some of the async races issues and makes the code simpler.

Do you have an example where this would be a disadvantage?

yasserf:
Let us assume we have a record with this content on the server:

```javascript
{
   firstName: "John",
   lastName: "Smith",
   country: "Berlin",
   street: "..."
}
```

A user would then say.. okay, I just loaded my application, setup my view, and would like to say person now lives in Oz. I'm going to be writing **pseudo code** based on KO, but the same applies to other things with functional aspects. 

```
const record = ds.record.getRecord( 'person' )

// does all bindings to dom here, lets say using KO
firstName = getObservable( record, 'name' )
lastName = getObservable( record, 'name' )
postalAddress = getComputedObservable( record, function( firstName, lastName, country, addess ) {
  return arguments.join( '\n' );
} )
```

Now if I was to set something like country here before whenReady

```
record.set( 'country', 'uk' )
```

My subscribers would be triggered, meaning my address ( previously empty since it wasn't yet initialised ) would contain a computed value of

```
undefined
undefined
uk
undefined
```

> When doing record.getRecord the user is already setting their local state to {} and setting the initial local state to something else should conceptually not matter.

I agree, and as you previously mentioned this is only an issue when doing PATCH, doing a full update makes sense since you throw away what was there/remote anyways. We have to differentiate the two, since users will get inconsistent data as in my previous example and the concept of patching/data-sync is new enough to be confusing without introducing more sync/async differentiation.

ronag:
@yasserf: I would say that the view code is broken then. I would have written it like. 

```javascript
const record = ds.record.getRecord( 'person' )
const firstName$ = record.observe('firstName')
const lastName$ = record.observe('lastName')
const country$ = record.observe('country')
const address$ = record.observe('address')
const postalAddress$ = Observable.combineLatest(firstName$, lastName$, country$, address$, ...)
```

Why would the user assume that all 4 of them exists? I feel like there is some form assumed implicit guarantee that we are trying to enforce. Which to me feels like an anti-pattern, i.e. make it work for bad code.

ronag:
At least in our code we never assume that a field cannot be `undefined` and if we have an assumption in regards to what is a "valid" state we check for that.

yasserf:
After discussing this with the core team we decided to not go ahead with these changes until we investigate offline behavior in general. 

This is mainly so we can investigate the other async issues in the client and concentrate on the roadmap for 2016

PR #222, opened 2016-08-23
Checks

ronag:
This makes debugging user code a lot easier. Especially since invalid arguments sometimes cause asynchronous errors, potentially crashing the client and giving confusing stack traces.

### DIFF
--- a/src/event/event-handler.js
+++ b/src/event/event-handler.js
@@ -31,19 +31,26 @@ var EventHandler = function( options, connection, client ) {
  * Subscribe to an event. This will receive both locally emitted events
  * as well as events emitted by other connected clients.
  *
- * @param   {String}   eventName
+ * @param   {String}   name
  * @param   {Function} callback
  *
  * @public
  * @returns {void}
  */
-EventHandler.prototype.subscribe = function( eventName, callback ) {
-	if( !this._emitter.hasListeners( eventName ) ) {
-		this._ackTimeoutRegistry.add( eventName, C.ACTIONS.SUBSCRIBE );
-		this._connection.sendMsg( C.TOPIC.EVENT, C.ACTIONS.SUBSCRIBE, [ eventName ] );
+EventHandler.prototype.subscribe = function( name, callback ) {
+	if ( typeof name !== 'string' || name.length === 0 ) {
+		throw new Error( 'invalid argument name' );
+	}
+	if ( typeof callback !== 'function' ) {
+		throw new Error( 'invalid argument callback' );
 	}
 
-	this._emitter.on( eventName, callback );
+	if( !this._emitter.hasListeners( name ) ) {
+		this._ackTimeoutRegistry.add( name, C.ACTIONS.SUBSCRIBE );
+		this._connection.sendMsg( C.TOPIC.EVENT, C.ACTIONS.SUBSCRIBE, [ name ] );
+	}
+
+	this._emitter.on( name, callback );
 };
 
 /**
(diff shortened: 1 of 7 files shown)


src/record/record-handler.js
```diff
@@ -273,6 +292,10 @@ RecordHandler.prototype._onRecordError = function( recordName, error ) {
  * @returns {void}
  */
 RecordHandler.prototype._onDestroyPending = function( recordName ) {
+	if ( !this._records[ recordName ] ) {
+		this.emit( 'error', 'Record \'' + recordName + '\' does not exists' );
```
yasserf: Exist*

src/record/record.js
```diff
@@ -234,6 +255,9 @@ Record.prototype.unsubscribe = function( pathOrCallback, callback ) {
  * @returns {void}
  */
 Record.prototype.discard = function() {
+	if( this._checkDestroyed( 'discard' ) ) {
```
yasserf: Should this not throw a already discarded error?

yasserf:
Really useful additions, can merge in after you see comments

src/record/record.js
ronag: Just following the convention used for the rest of the code... not sure how you would like to have it?

src/event/event-handler.js
```diff
@@ -31,38 +31,51 @@ var EventHandler = function( options, connection, client ) {
...
-	this._emitter.off( eventName, callback );
-	
-	if( !this._emitter.hasListeners( eventName ) ) {
-		this._ackTimeoutRegistry.add( eventName, C.ACTIONS.UNSUBSCRIBE );
-		this._connection.sendMsg( C.TOPIC.EVENT, C.ACTIONS.UNSUBSCRIBE, [ eventName ] );
+EventHandler.prototype.unsubscribe = function( name, callback ) {
+	if ( typeof name !== 'string' || name.length === 0 ) {
+		throw new Error( 'invalid argument name' );
+	}
+	if ( callback !== undefined && typeof callback !== 'function' ) {
```
ronag: Not sure if `callback === undefined` should be allowed?

ronag:
Not I added check for rpc and event as well.

src/event/event-handler.js
yasserf: that should then remove all callbacks with that name, so should be allowed up

src/rpc/rpc-response.js
```diff
@@ -11,6 +11,10 @@ var C = require( '../constants/constants' ),
  * @param {String} correlationId the correlationId for the RPC
  */
 var RpcResponse = function( connection, name, correlationId ) {
+	if ( typeof name !== 'string' || name.length === 0 ) {
```
yasserf: this is generated internally, shouldn't need a check

src/rpc/rpc.js
```diff
@@ -14,6 +14,10 @@ var C = require( '../constants/constants' ),
  * @constructor
  */
 var Rpc = function( options, callback, client ) {
+	if ( typeof callback !== 'function' ) {
```
yasserf: this is generated internally, shouldn't need a check

src/record/record.js
yasserf: Yup, check destroyed emits an error.

```
Record.prototype._checkDestroyed = function( methodName ) {
	if( this.isDestroyed ) {
		this.emit( 'error', 'Can\'t invoke \'' + methodName + '\'. Record \'' + this.name + '\' is already destroyed' );
		return true;
	}

	return false;
};
```

yasserf:
This is generally great! I'll merge it in today!

PR #223, opened 2016-08-24
Handle exceptions thrown by user callback

ronag:
Currently it is assumed that any callback sent in by the user NEVER throws. If it does throw then the deepstream state is pretty much undefined.

This is WIP.

### DIFF
--- a/src/record/record.js
+++ b/src/record/record.js
@@ -272,8 +272,10 @@ Record.prototype.delete = function() {
  */
 Record.prototype.whenReady = function( callback ) {
 	if( this.isReady === true ) {
+		// TODO: What if callback throws?
 		callback( this );
 	} else {
+		// TODO: What if callback throws?
 		this.once( 'ready', callback.bind( this, this ) );
 	}
 };
@@ -329,6 +331,7 @@ Record.prototype._recoverRecord = function( remoteVersion, remoteData, message )
 		this._mergeStrategy( this, remoteData, remoteVersion, this._onRecordRecovered.bind( this, remoteVersion ) );
 	}
 	else {
+		// TODO: What if callback throws?
 		this.emit( 'error', C.EVENT.VERSION_EXISTS, 'received update for ' + remoteVersion + ' but version is ' + this.version );
 	}
 };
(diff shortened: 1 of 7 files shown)


ronag:
For now I've just identified potential problem areas.

@yasserf: How should this be handled, especially in the asynchronous case? `client.emit( 'error', ...)`?

src/record/record.js
```diff
@@ -453,9 +459,11 @@ Record.prototype._onRead = function( message ) {
 Record.prototype._setReady = function() {
 	this.isReady = true;
 	for( var i = 0; i < this._queuedMethodCalls.length; i++ ) {
+		// TODO: What if callback throws?
 		this[ this._queuedMethodCalls[ i ].method ].apply( this, this._queuedMethodCalls[ i ].args );
```
ronag: This will never throw if we properly validate all the arguments in `set` before enqueueing it.

ronag: With #221 this wouldn't be a problem.

ronag:
```javascript
tryCatchEmit(client, callback) {
  try {
    callback();
  } catch ( err ) {
    client.emit( 'error', /* how should this look? */);
  }
}
```

ronag:
What I'm mostly unsure about is how to handle the `emit` callbacks. Not sure how NodeJS resolves this issue.

ronag:
It seems the practice in Node is to try/catch the emit and let the user shoot themselves if they have a throwing listeners, in terms of having all listeners called.

yasserf:
do you mean we put the try catch around the for loop then, and if one of them fail it just doesn't notify the rest for that event?

ronag:
@yasserf: That seems to be the practice for internal NodeJS code. However, we should still do `client.emit('error', 'uncaught exception')` or something.

ronag:
If we want to catch for each callback we would have to modify `component-emitter`.

yasserf:
yeah that would be way too much of a performance hit, I like the idea of wrapping the loop, but that in itself is also expensive if ( like in your case ) you have records being requested and discarded constantly

ronag:
>but that in itself is also expensive

I would prefer reliable state over performance in this case.

ronag:
@yasserf: Do you know of any interesting optimisation from V8 in regards to try/catch the last year?

yasserf:
No, the last time I checked this still makes the method unoptimized

AlexBHarley:
Closing due to PR age

PR #233, opened 2016-09-02
Feature/presence

AlexBHarley:


src/presence/presence-handler.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
```
yasserf: client can't use Class syntax since we support older browsers

src/presence/presence-handler.js
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	getCurrentClients( callback ) {
+		this._presentClientsCallback = callback;
+		var msg = messageBuilder.getMsg( C.TOPIC.PRESENCE, C.ACTIONS.QUERY );
+		this._connection.send( msg );
+	}
+
+	onClientLogin( callback ) {
+		this._clientAddedCallback = callback;
```
yasserf: we would want to use an event emitter and do an `on` here and `off` somewhere. 

We should look at the API, since it might be better to do a 

`on( 'clientLoggedOn', () => {} )`

aware this is just first commit though!=D

test-e2e/specs/presenceSpec.js
```diff
@@ -0,0 +1,95 @@
+/* global describe, it, expect, jasmine */
+var DeepstreamServer = require( '../../../deepstream.io/src/deepstream.io' ),
```
AlexBHarley: assumes folders are next to eachother, ie:
`- dev`
`-- deepstream.io`
`-- deepstream.io-client-js`

src/client.js
```diff
@@ -114,6 +118,19 @@ Client.prototype.getUid = function() {
 	return timestamp + '-' + randomString;
 };
 
+Client.prototype.getCurrentClients = function( callback ) {
+	this.presence.getCurrentClients( callback );
```
yasserf: can you return this for chaining?

src/presence/presence-handler.js
```diff
@@ -0,0 +1,49 @@
...
+		this._options = options;
+		this._connection = connection;
+		this._client = client;
+		this._emitter = new EventEmitter();
+		this._ackTimeoutRegistry = new AckTimeoutRegistry( client, C.TOPIC.PRESENCE, this._options.subscriptionTimeout );
+};
+
+PresenceHandler.prototype.getCurrentClients = function( callback ) {
+	this._emitter.on( C.ACTIONS.QUERY, callback );
+	this._presentClientsCallback = callback;
```
yasserf: no longer needed

test-e2e/specs/presenceSpec.js
yasserf: could just do a `require( 'deepstream.io` ) and install deepstream via 

```
npm install https://github.com/deepstreamIO/deepstream.io.git
```

yasserf: on your local machine you could also do:

```
cd deepstream.io && npm link
```
then
```
cd deepstream.io-client-js && npm link deepstream.io
```

test-e2e/specs/presenceSpec.js
```diff
@@ -0,0 +1,95 @@
+/* global describe, it, expect, jasmine */
```
yasserf: you are going to totally not like me for this, but we are now writing our [specs](https://github.com/deepstreamIO/deepstream.io-e2e)  in cucumber (a)

On the bright side, its ALOT easier in jsland! I'm more than happy to help out with that if needed.

yasserf: Idea is so that we can share the same tests across multiple clients, and the framework takes care of setting up a cluster and everything for you so u can concentrate on just writing steps for whatever it is instead

test-unit/unit/presence/presence-handlerSpec.js
```diff
@@ -0,0 +1,78 @@
+/* global describe, expect, it, jasmine */
```
yasserf: looks awesome!

PR #234, opened 2016-09-02
Fix/#229

### LINKED ISSUE #229: subscription callback can be called even if unsubscribed
Consider the following:

``` javascript
const rec = ds.record.getRecord('foo')
const doNotCall = () =>() => console.log('NO')
rec.subscribe(doNotCall, true);
rec.unsubscribe(doNotCall);
// oops... NO
```

The problem is a combination of `triggerNow` and `whenReady`.

ronag:
Based on #214. Fixes #229.

I've changed the subscriptionSpec tests. Initial read should invoke registered subscriptions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,3 @@ node_js:
   - "stable"
 
 script: npm run coverage
-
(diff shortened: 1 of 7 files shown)


yasserf:
Closing after re-base on master -> #240

PR #240, opened 2016-09-04
#229

yasserf:
Based on #229

### DIFF
--- a/src/record/record.js
+++ b/src/record/record.js
@@ -169,13 +169,10 @@ Record.prototype.subscribe = function( path, callback, triggerNow ) {
 		return;
 	}
 
-	if( args.triggerNow ) {
-		this.whenReady( function () {
-			this._eventEmitter.on( args.path, args.callback );
-			args.callback( this.get( args.path ) );
-		}.bind(this) );
-	} else {
-		this._eventEmitter.on( args.path, args.callback );
+	this._eventEmitter.on( args.path, args.callback );
+
+	if( args.triggerNow && this.isReady ) {
+		args.callback( this.get( args.path ) );
 	}
 };
 
@@ -449,7 +446,16 @@ Record.prototype._applyUpdate = function( message ) {
  */
 Record.prototype._onRead = function( message ) {
 	this.version = parseInt( message.data[ 1 ], 10 );
-	this._applyChange( jsonPath.set( this._$data, undefined, JSON.parse( message.data[ 2 ] ) ) );
+
+	this._$data = JSON.parse( message.data[ 2 ] );
+	if ( this._eventEmitter._callbacks ) {
+		var paths = Object.keys( this._eventEmitter._callbacks );
+
+		for ( var i = 0; i < paths.length; i++ ) {
+			this._eventEmitter.emit( paths[ i ], this.get( paths[ i ] ) );
+		}
+	}
+
 	this._setReady();
 };
 
--- a/test-unit/unit/record/record-subscriptionSpec.js
+++ b/test-unit/unit/record/record-subscriptionSpec.js
@@ -16,7 +16,7 @@ describe( 'supscriptions to local record changes', function(){
 		record.subscribe( anotherGeneralCallback );
 		expect( generalCallback ).not.toHaveBeenCalled();
 		record._$onMessage({ topic: 'RECORD', action: 'R', data: [ 'testRecord', 0, '{}' ]} );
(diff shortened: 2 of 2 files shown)


src/record/record.js
```diff
@@ -449,7 +446,16 @@ Record.prototype._applyUpdate = function( message ) {
  */
 Record.prototype._onRead = function( message ) {
 	this.version = parseInt( message.data[ 1 ], 10 );
-	this._applyChange( jsonPath.set( this._$data, undefined, JSON.parse( message.data[ 2 ] ) ) );
+
+	this._$data = JSON.parse( message.data[ 2 ] );
```
yasserf: @ronag is the change from `_applyChange` to doing it all inline an optimisation? Using applyChange would stop it from triggering undefined on the all the paths that don't exist yet on the first update.

yasserf: Which is what we done before in the previous method, apologies, ignore that comment.

yasserf: Okay, so:

###### Old behavior - record is/isn't ready - triggerNow true:
1) Subscribe to path `name`
2) Get read -> {}
3) name triggered with undefined


###### Old behavior - record is/isn't ready - triggerNow true:
1) Subscribe to path `name`
2) Get read -> {}
3) name triggered with undefined
###### New behavior - record is ready:
1) Subscribe to path `name`
2) Nothing..

###### New behavior - record isn't ready:
1) Subscribe to path `name`
2) Get read -> {}
3) name triggered with undefined

ronag: Hm... yes the problem with the new behaviour is that non-existing paths would be triggered.

ronag: The problem with the old behaviour is that:

1) Subscribe to path name
2) Get read -> {}
3) name triggered with undefined, THIS IS NOT TRUE?

src/record/record.js
```diff
@@ -449,7 +446,16 @@ Record.prototype._applyUpdate = function( message ) {
...
+	if ( this._eventEmitter._callbacks ) {
+		var paths = Object.keys( this._eventEmitter._callbacks );
+
+		for ( var i = 0; i < paths.length; i++ ) {
+			this._eventEmitter.emit( paths[ i ], this.get( paths[ i ] ) );
```
ronag: This should be:
```javascript
``` if ( jsonPath.get( this._$data, paths[ i ], false) !== undefined)
```

src/record/record.js
yasserf: Okay, so cases ontop of my head:

###### Old behavior - record is/isn't ready - triggerNow true:
1) Subscribe to path `name`
2) Get read -> {}
3) name triggered with undefined

###### Old behavior - record is/isn't ready - triggerNow false:
1) Subscribe to path `name`
2) Get read -> {}
3) nothing...

###### New behavior - record is ready - triggerNow false:
1) Subscribe to path `name`
2) Nothing..

###### New behavior - record isnt ready - triggerNow false:
1) Subscribe to path `name`
2) Get read -> {}
3) name triggered with undefined

###### New behavior - record is ready - triggerNow true:
1) Subscribe to path `name`
3) name triggered with undefined

###### New behavior - record isnt ready - triggerNow true:
1) Subscribe to path `name`
2) name triggered with undefined
3) Get read -> {}
4) name triggered with undefined

ronag: We can't use `applyChange` here since it would not trigger if the read value is `{}`.

ronag: Your last example is incorrect:

1) Subscribe to path name
**2) name triggered with undefined** Nope
3) Get read -> {}
4) name triggered with undefined

yasserf: Good point, updated

src/record/record.js
yasserf: That was my point with my previous comment ( which I should have spent a bit more time on! ),

Before we always updated all initial paths with `triggerNow=true`, regardless if undefined or not. 
If `triggerNow=false` we never notified paths when undefined.

The new change either notifies everything with undefined if `triggerNow=true` and `isReady=false`, or with the previous comment, will never notify anyone if is `triggerNow=false`

Generally this was an inconsistency in the code previously, I'm just wondering if:
a) It creates any compatibility issues for those not using `isSubscribed`
b) If telling people it is undefined initially is desired, as some might take it as an indication to actually populate missing fields.

Will need to discuss with team to see what the general agreement is. Otherwise, this fixes the bug #229 well!

yasserf:
Closing this because of how outdated it currently is / merge issues

Issue #266, opened 2016-10-24
Websocket error in v2.0


ronag:
```
07:15:34.346Z ERROR asset: Deepstream Error. (error="Two connections heartbeats missed successively", event=connectionError, topic=C)
node_modules/ws/lib/WebSocket.js:219
    else throw new Error('not opened');
         ^

Error: not opened
    at WebSocket.send (node_modules/ws/lib/WebSocket.js:219:16)
    at Connection._handleConnectionResponse (node_modules/deepstream.io-client-js/src/message/connection.js:386:18)
    at Connection._onMessage (node_modules/deepstream.io-client-js/src/message/connection.js:349:9)
    at WebSocket.onMessage (node_modules/ws/lib/WebSocket.js:442:14)
    at emitTwo (events.js:106:13)
    at WebSocket.emit (events.js:191:7)
    at Receiver.ontext (node_modules/ws/lib/WebSocket.js:841:10)
    at node_modules/ws/lib/Receiver.js:536:18
    at Receiver.applyExtensions (node_modules/ws/lib/Receiver.js:371:5)
    at node_modules/ws/lib/Receiver.js:508:14
[nodemon] app crashed - waiting for file changes before starting...
```

yasserf:
I saw this issue in ws in a few of our test frameworks, I tracked it down to an odd issue with ws not propagating the close event before changing its state to closed.

When you raised issues against master/unreleased versions can you please mention the version? Having an issue just titled  `crashes` sounds a bit intense.

ronag:
@yasserf: Yes sorry about that. Was in a hurry.

ronag:
@yasserf: Is this something that needs to be fixed in WS or DS?

yasserf:
Not entirely sure yet tbh, I have a fix in DS that checks for the connection state on the websocket but is a workaround.

ronag:
@yasserf: send can throw. https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#Exceptions_thrown_2

yasserf:
send can throw when the connection state is not open, which is something we should get notified about via ws.

A better solution would be to check if the connection ready state is open rather than via try catch

```
if (this._endpoint.readyState !== WebSocket.OPEN) {
 // do something like log an error or queue message
}
```

ronag:
> which is something we should get notified about via ws.

Are you sure it has to notify synchronously before you are able to call send?

yasserf:
[...]

PR #273, opened 2016-11-01
Bugfix/#266 Websockets trying to send while not open

jdmnd:
https://github.com/deepstreamIO/deepstream.io-client-js/issues/266


ronag:
Needs a check at https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/message/connection.js#L374 as well.

Issue #256, opened 2016-10-07
Typescript typings


jensvdh:
Does this library have Typescript typings? If not I'd love to contribute to this project by building them and opening a PR.

yasserf:
It currently does not, but that would be awesome if you do!

jugglingcats:
@jensvdh did you get anywhere with this?

jensvdh:
@jugglingcats I have gotten a bit busy with work. But I should have a PR with Typescript typings up over the next week.

stemcc:
@jugglingcats You could always gist what you've got so far and perhaps the community can take it the rest of the way.

RichiWIP:
I have created a quick d.ts to use in a current project of mine. Hope this helps.
https://gist.github.com/RichiWIP/691235f5257e014b3e8a040dfe247187

jensvdh:
Hi @RichiWIP. Awesome work! I just tested the definitions locally and they seem to work great for the most part. Although the presence API seems to be missing.

I went ahead and created a pull request here

https://github.com/deepstreamIO/deepstream.io-client-js/pull/283

This should bundle the typings with this repository, making it near automatic for TS developers to get started.

benjaminVadon:
Hi all,
I'm pretty a new to typescript and I'm attempting to do something like that using the d.ts file provided by @RichiWIP 
```
client.on("connectionStateChanged", function (state: ConnectionState) {
            switch (state) {
                case ConnectionState.OPEN:
                    console.log("open");
                    break;
                default:
                    console.log("not open");
            }
      })
```

Doing so, webstorm suggests me to import enum this way
`import ConnectionState = deepstreamIO.ConnectionState;`

In typescript there is no problem, enum is well recognize, but after bundled all in JS file, and then open it  with chrome, it complains that 
> VM396:1 Uncaught ReferenceError: deepstreamIO is not defined

I've attempted many differents things (other kind of import, adding export in enum, change deepstream export object), probably insane for a typescript pro dev, that I'm not yet.

I've already searched in ts documentation, github, google about namespace, module, enum, ambient context,... and can't find a way to achieved that.
So, is it possible to use enum declared d.ts ? and if yes, can you help me or drive to the right place to learn ?

jensvdh:
Hi.

There's a pull request open for the Typescript definitions.

Can you try installing this specific branch from NPM until it's been merged in?

https://github.com/deepstreamIO/deepstream.io-client-js/pull/283

[...]

PR #283, opened 2016-11-29
Added Typescript declarations.

jensvdh:
Hi,

I added Typescript definitions to this repository. This should help Typescript developers a lot by providing type declarations for this repository.

See issue #256 

Big thanks to @RichiWIP whom provided the first version of a Typescript declaration file.

I decided to follow Microsoft's recommended new guidelines that suggest bundling the Typescript Declarations with the module and adding a reference to package.json. Therefore developers won't have to download the typescript definitions from an external repository.

see:
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

Alternatively we can create a second repository and submit our typings to Microsoft's @types project directly. However, this process is more involved and no longer seems recommended.

I added some steps to readme.md to get Typescript developers started. Let me know if you'd like to put this documentation somewhere else.

### DIFF
--- a/src/client.d.ts
+++ b/src/client.d.ts
@@ -1,0 +1,294 @@
+declare namespace deepstreamIO {
+
+        enum ConnectionState {
+                CLOSED, AWAITING_CONNECTION, CHALLENGING, AWAITING_AUTHENTICATION, AUTHENTICATING, OPEN, ERROR, RECONNECTING
+        }
+
+        enum Event {
+                TOO_MANY_AUTH_ATTEMPTS, NOT_AUTHENTICATED, CONNECTION_ERROR, MESSAGE_PERMISSION_ERROR, MESSAGE_PARSE_ERROR, MESSAGE_DENIED, LISTENER_EXISTS, NOT_LISTENING, IS_CLOSED, ACK_TIMEOUT, RESPONSE_TIMEOUT, DELETE_TIMEOUT, UNSOLICITED_MESSAGE, RECORD_NOT_FOUND, VERSION_EXISTS, UNKNOWN_CALLEE
+        }
+
+        enum MergeStrategy {
+
+        }
+
+        interface Options {
+                /**
+                 *A global merge strategy that is applied whenever two clients write to the same record at the same time. Can be overwritten on a per record level. Default merge strategies are exposed by the client constructor. It's also possible to write custom merge strategies as functions. You can find more on handling data conflicts here
+        Type: Function
+        Default: MERGE_STRATEGIES.REMOTE_WINS
+                 */
+                mergeStrategy?: MergeStrategy;
+                /**Specifies the number of milliseconds by which the time until the next reconnection attempt will be incremented after every unsuccessful attempt.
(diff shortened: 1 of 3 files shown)


WolframHempel:
Awesome, thank you very much. We'll review it and merge it shortly

src/client.d.ts
```diff
@@ -0,0 +1,294 @@
...
+        Info
+        
+        Requests count as completed once send() or error() was called. Calling ack() after that won't do anything.
+                 */
+                ack(): void;
+        }
+
+        interface RecordStatic {
+                /** Get a record. */
+                getRecord(path: string): Record;
```
yasserf: getAnonymousRecord is missing here ( as well as an interface )

yasserf:
Any chance anyone tested or used this? Be great to get in for 2.1

jensvdh:
What's the schedule for 2.1? We noticed a couple of issues  that need to be addressed with the typings.

yasserf:
We are aiming to get it out by early next week. No rush though, we can easily just merge it in when your happy and do a separate release, our release cycle is mostly automated ;) =D

NathanFlurry:
I started trying out this declaration file and immediately ran into a few issues. First of all, none of the interfaces or enums are exported, so it's impossible to store values like the `Client` in TypeScript as a non-`any` type. Additionally, TypeScript enums are not the same as string enums. For instance, `ConnectionState.OPEN` would equal `5` based on the declaration file, but in the JavaScript library, it equals the literal string, `"OPEN"`. Therefore, I believe it needs a bit more work. If I get a chance, I'll look into fixing it.

jensvdh:
I believe you can use strings in ENUM's now in Typescript like this:
https://blog.rsuter.com/how-to-implement-an-enum-with-string-values-in-typescript/

yasserf:
I'm sorry for the delay on merging this, we are still trying to figure out the best way of testing and maintaining it going forward. It will be merged, just pending the next client milestone.

src/client.d.ts
```diff
@@ -0,0 +1,294 @@
...
+
+                rpc: RPCStatic;
+        }
+
+        interface deepstreamQuarantine extends Client {
+
+        }
+        interface deepstreamStatic {
+                CONSTANTS: {
+                        [key: string]: string
```
jensvdh: That works. We'll also want to update the enums to use strings. Typescript enums compile to integers.

yasserf:
@jensvdh @benjaminVadon Are you both okay with me merging this?

jensvdh:
I'm OK with merging this. It's not perfect yet but we can improve on it in a 2nd PR. Have these typings is still an improvement over having none.

yasserf:
Agreed!

yasserf:
One last thing, would you be okay with signing the CLA? https://deepstream.io/info/community/cla/

I have said that a bit often lately so if you already have my bad!

jensvdh:
Done!

yasserf:
Awesome, and thanks again for raising this!

PR #284, opened 2016-11-29
Presence spec tests

AlexBHarley:


test-specs/steps/client/presence.js
```diff
@@ -0,0 +1,49 @@
...
+		setTimeout( callback, config.messageWaitTime );
+	});
+
+	this.When( /^the client unsubscribes to presence events$/, function( callback ){
+		global.dsClient.presence.unsubscribe( subscribeCallback );
+		setTimeout( callback, config.messageWaitTime );
+	});
+
+	this.When( /^the client is notified that client "(\w*)" logged in$/, function( username ){
+		sinon.assert.calledWith( subscribeCallback, username, true );
```
yasserf: its also good practice to assert they have only been called once, since we treat these more as mocks than spies ( explicitly note each api call )

PR #287, opened 2016-12-01
fix/json-path - adding null check

### LINKED ISSUE #281: json-path -> patch: TypeError: Cannot read property of null
Hello,
This happens in `v2.0.0`, but not in `1.0.0` (I discovered the issue after upgrading). The culprit seems to be the `patch` function in the `record/json-path.js` module, and the issue seems to have been introduced in [this commit](https://github.com/deepstreamIO/deepstream.io-client-js/commit/f687485a420b77e9d1922e6032f4b3524dab042b).

My particular case is kinda complex, but debugging the client it seems that this exception gets thrown when `patch` is called by  `set` and a value lower in the tree goes from falsy (or possibly even non-object) to object. 

In my case, `oldValue` is descended recursively in patch, until it gets to a sub-object of the form:

**oldValue**
```javascript
{
  job: null,
  joinedAt: 1480020987915
}
```

**newValue**
```javascript
{
  job: { id: 88 },
  joinedAt: 1480020987915
}
```

**deepCopy**
```javascript
true
```

At this point, `patch` is recursively called with the `job` sub-object, and the arguments are:

**oldValue**
```javascript
null
```

**newValue**
```javascript
{ id: 88 }
```

**deepCopy**
```javascript
true
```

This makes the client crash with an exception, when [this line](https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/record/json-path.js#L97) is executed:

```javascript
obj[ props[ i ] ] = patch( oldValue[ props[ i ] ], newValue[ props[ i ] ], deepCopy )
```

Since `props` is the keys of `newValue`, this will try to compute `oldValue.id`. But `oldValue` is `null` and an exception gets thrown. It seems that an additional check for this (pretty common) case is needed in `patch`.

AlexBHarley:
Fixes #281 - unit tests to follow

### DIFF
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -83,6 +83,9 @@ function patch( oldValue, newValue, deepCopy ) {
 	if ( utils.deepEquals( oldValue, newValue ) ) {
 		return oldValue;
 	}
+	else if ( oldValue === null || newValue === null ) {
+		return newValue;
+	}
 	else if ( Array.isArray( oldValue ) && Array.isArray( newValue ) ) {
 		var arr = [];
 		for ( i = 0; i < newValue.length; i++ ) {
--- a/test-unit/unit/record/json-pathSpec.js
+++ b/test-unit/unit/record/json-pathSpec.js
@@ -113,6 +113,20 @@ describe( 'objects are created from paths and their value is set correctly', fun
 		});
 	});
 
+	it( 'sets values for null values', function(){
+		var record = { _$data:{
+			job: null,
+  			joinedAt: 1480020987915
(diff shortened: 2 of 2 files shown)


src/record/json-path.js
```diff
@@ -83,6 +83,9 @@ function patch( oldValue, newValue, deepCopy ) {
 	if ( utils.deepEquals( oldValue, newValue ) ) {
 		return oldValue;
 	}
+	else if ( oldValue === null || newValue === null ) {
```
yasserf: No related test? (A)

PR #295, opened 2016-12-14
Resolves #289

### LINKED ISSUE #289: Feature Request - Pass options to node websocket
I use self signed certs when developing locally, and I need a way of setting the `rejectUnauthorized` option to false. I would prefer this to disabling such checks globally.

Even setting the `ca` option could be helpful for production use.

I am willing to make a PR, but do not know how you all would like the options to be passed.

```javascript
deepstream(url, options);

/**
 * Where options could contain one of the following:
 * options.socket: Object
 * options.socketOptions: Object
 * 
 * or even perhaps:
 * options.rejectUnauthorized: Boolean
 * options.ca: String|String[]|Buffer|Buffer[]
 */
```

**Relavent Links**
- Deepstream's use of sockets at [connection.js#L151](https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/message/connection.js#L151)
- WebSocket API can be seen at [/ws/WebSocket.js#L36](https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L36).
- WebSocket list of options can be seen at [/ws/WebSocket.js#L386](https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L386).

Lalem001:
Resolves #289 

I opted for the parameter name `nodeSocketOptions`.

I am not sure how to adjust tests for this since it simply passes straight through to the websocket constructor.

**Changes**

- Allows configuration to pass options to websocket constructor in node

### DIFF
--- a/src/default-options.js
+++ b/src/default-options.js
@@ -112,5 +112,12 @@ module.exports = {
 	 *                                  improves speed at the expense of the user having to ensure
 	 *                                  object immutability.
 	 */
-	recordDeepCopy: true
+	recordDeepCopy: true,
+
+	/**
+	 * @param {Object} nodeSocketOptions    Options to pass to the websocket constructor in node.
+	 * @default null
+	 * @see https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketaddress-protocols-options
+	 */
+	nodeSocketOptions: null
 };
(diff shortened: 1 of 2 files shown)


yasserf:
Looks good, thanks!

Lalem001:
Any idea when this could be published to npm?

PR #308, opened 2017-01-08
Rewritten in TypeScript

NathanFlurry:
Basically, I was working on updating the declaration file for PR #283 and  somehow 8 hours later I ended up rewriting the entire project in TypeScript. And here it is in all its glory.

Notable changes for the API consumers:
- This is now using the ES6 module declaration, so instead of exporting one function, `createDeepstream`, now everything is exported so you can create a client like `new Client(...)`.
  - For instance, to import a constant, the syntax is now something like `import { Actions } from "../constants/Constants";` instead of `import C = require("../constants/Constants")`
  - Because of this, the project will now work properly in TypeScript, since the declarations for `Client` and everything else is exported, so it can be stored in a typed variable. Before, to store the client, one had to declare a variable like `let x: any;` then do `x = new Client(...)`; this removed all inline suggestions, which was annoying.
- Some spelling errors were fixed in API methods.
- Constants were renamed to appear more elegant.
- `Client._options` is not passed to every object now. When the options are created on construction, the object is frozen and exposed publicly with a getter. Then, every other internal class defines their own getter like `get _options(): DeepstreamOptions { this._client.options; }`
- When appropriate, `Promises` were put in place of callbacks.

Anyways, Deepstream is an amazing project and it's going to be around for a long time, so the library needs to move away from ES5 at some point and TypeScript helps make it more or less "future proof". Additionally, the strict typing is very handy for catching errors in such a complex project.

I understand that this update makes some breaking changes to the library, so this is probably something that should be merged for a major release.

Thanks,
Nathan Flurry

P.S. It's really interesting to see how errors that weren't caught by the TypeScript compiler and not the unit tests.


yasserf:
Hey Nathan,

I'm super impressed, this was quite an undertaking! Unfortunately it's way to big of a change for us to merge in, and figure out with tests, compilation and so forth. 

I'm going to close down this PR in favour of [this issue](https://github.com/deepstreamIO/deepstream.io-client-js/issues/309) 

Thanks!

yasserf:
hey Nathan,

Did you add a step that transpiled those to ie9+ compatability? We are having an internal discussion of using typescripts instead of babel when going forward.

Cheers!

NathanFlurry:
@yasserf Right now, I believe it's just targeting ES6 in the TypeScript configuration file and leaving the compilation to ES5 + old browsers up to Grunt/Babel. I only tested it in NodeJS, but I can mess with it later today to try and get it working in the browser – it shouldn't be difficult at all.

That said, in terms of the build systems, I've found using Webpack 2 + ts-loader works by far the best for TypeScript. In my experience, TypeScript's built in support for compiling to a single file (e.g. `compilerOptions.outFile` in the `tsconfig.json`) doesn't work very well and is prone to errors and headaches.

PR #316, opened 2017-01-26
index is undefined when record is not ready

ralphtheninja:
If the record isn't in a ready state, calls to `list.removeEntry()` and `list.addEntry()` will not work since the queued methods will always ignore the provided index.

### DIFF
--- a/src/record/list.js
+++ b/src/record/list.js
@@ -103,7 +103,7 @@ List.prototype.setEntries = function( entries ) {
 /**
  * Removes an entry from the list
  *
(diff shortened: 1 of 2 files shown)


ralphtheninja:
I was really going to fix callbacks for lists, but I needed the optional index to be passed on to `.bind()` before doing that.

ralphtheninja:
@yasserf Any objection to merging this?

yasserf:
@ralphtheninja  None no, merging in now!

Issue #327, opened 2017-02-16
record.set() fails to set some keys


lulzmachine:
Hello!

We ran into an issue recently when building a record. We wanted to have a record containing a bunch of mac addresses. We wanted to use the mac addresses as keys. But we ran into this problem:

```
var macsRecord = ds.record.getRecord('some/path').whenReady(record => {
    macsRecord.set('0x000347971', 'macrecord')
    console.log(macsRecord.get())

    // will output { '0': 'macrecord' }

    macsRecord.set({
      '0x000347972': 'macrecord2'
    })
    console.log(macsRecord.get())

   // will output { '0x000347972': 'macrecord2' }
})
```

The expected result is that both methods produce the same output (like the latter method). I.e. it shouldn't turn a valid key string into a string containing just a zero.

The problem seems to be in the "tokenize" method in json-path.js:
```
	return cache[ path ] = parts.map( function( part ) {
		return !isNaN( part ) ? parseInt( part, 10 ) : part;
	} );
```

For some reason it's running "parseInt(part, 10)" on the key, which will turn it into 0.

We're running "deepstream.io-client-js": "^2.1.1"

lulzmachine:
I guess it's meant to do parseInt() for things that are base-10-numbers. Unfortunately javascript accepts things as !isNaN than decimal numbers.
FYI:
```
> isNaN('0x23')
false
> isNaN('0f23')
true
> parseInt('0x234', 16)
564
> parseInt('0x234', 10)
0
> parseInt('0x234')
564
```

I guess the solution could either be to remove the "10" part from parseInt, or to just remove this whole thing altogether. I can't find any indication in the tests or the code why this tokenization exists.

ralphtheninja:
So maybe we could replace `parseInt(part, 10)` with just `parseInt(part)`? It should work out of the box for base 10 numbers and also fix the hex problem.

ralphtheninja:
@yasserf @WolframHempel 

What is the reason behind this parseInt-conversion in the first place? I suspect some kind of optimization of key sizes. Would it be a terrible solution to remove it altogether? :) Alternatively, a way to opt-out of the optimization and store the keys as-is.

lulzmachine:
I tried how the code works when changing parseInt(part, 10) to parseInt(part):

```
    macsRecord.set('0x000347971', 'macrecord')
    console.log(macsRecord.get())
// { '3438961': 'macrecord }
[...]

PR #328, opened 2017-02-16
Added support for non-NaNish base 16 numbers in jsonpath

lulzmachine:
This is related to https://github.com/deepstreamIO/deepstream.io-client-js/issues/327

### DIFF
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -124,6 +124,6 @@ function tokenize( path ) {
 	}
 
 	return cache[ path ] = parts.map( function( part ) {
-		return !isNaN( part ) ? parseInt( part, 10 ) : part;
+		return part;
 	} );
 };
--- a/src/record/json-path.js
+++ b/src/record/json-path.js
@@ -123,7 +123,5 @@ function tokenize( path ) {
 		throw new Error('invalid path ' + path)
 	}
 
-	return cache[ path ] = parts.map( function( part ) {
-		return part;
-	} );
+	return cache[ path ] = parts;
 };
--- a/test-unit/unit/record/json-pathSpec.js
+++ b/test-unit/unit/record/json-pathSpec.js
@@ -139,7 +139,7 @@ describe( 'objects are created from paths and their value is set correctly', fun
 
 	it( 'even when the path is not NaNish and could be interpreted as a base 16 number', function() {
 		var record = {};
-		let pathName = '0x02335';
+		var pathName = '0x02335';
 		record = jsonPath.set( record, pathName, 'value' );
 		expect( record[ 0 ] ).toBe( undefined );
 		expect( record[ pathName ] ).toBe( 'value' );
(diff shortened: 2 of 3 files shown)


yasserf:
Hey, awesome thank you! Would you mind also signing our CLA? https://deepstream.io/info/community/cla/

test-unit/unit/record/json-pathSpec.js
```diff
@@ -137,6 +137,16 @@ describe( 'objects are created from paths and their value is set correctly', fun
...
 	});
 
+	it( 'even when the path is not NaNish and could be interpreted as a base 16 number', function() {
+		var record = {};
+    let pathName = '0x02335';
+    record = jsonPath.set( record, pathName, 'value' );
+    expect( record[ 0 ] ).toBe( undefined );
+		expect( record[ pathName] ).toBe( 'value' );
+    expect( jsonPath.get( record, pathName )).toBe( 'value' );
+    expect( record[ pathName ]).toBe( 'value' );
```
ralphtheninja: Maybe adjust the tabs here :)

src/record/json-path.js
```diff
@@ -124,6 +124,6 @@ function tokenize( path ) {
 	}
 
 	return cache[ path ] = parts.map( function( part ) {
-		return !isNaN( part ) ? parseInt( part, 10 ) : part;
+		return part;
```
ralphtheninja: Maybe do

```js
return cache[ path ] = parts.slice(0);
```

lulzmachine: Hm actually, parts is created just a few lines above it, and isn't sent anywhere. So I guess I don't even need to copy it. I'll just `cache[ path ] = parts`

ralphtheninja: > Hm actually, parts is created just a few lines above it, and isn't sent anywhere. So I guess I don't even need to copy it. I'll just cache[ path ] = parts

Even better :)

test-unit/unit/record/json-pathSpec.js
```diff
@@ -137,6 +137,16 @@ describe( 'objects are created from paths and their value is set correctly', fun
 		});
 	});
 
+	it( 'even when the path is not NaNish and could be interpreted as a base 16 number', function() {
+		var record = {};
+		let pathName = '0x02335';
```
yasserf: let fails build ( we can put a use strict at the top since this just tests though if you want )

lulzmachine: Ok. I changed to "var". Thanks!

yasserf:
Great thanks!

ralphtheninja:
@yasserf How often do you release versions? We could really use an updated version of the client lib. I'm not sure what to bump here, maybe it should be `npm version major` since this PR affects the key format.

yasserf:
Hey @ralphtheninja, can do a quick release with that. Don't think this should be a major though, otherwise we will need to postpone it till 3.0 which is not due for a while.

ralphtheninja:
> Hey @ralphtheninja, can do a quick release with that. Don't think this should be a major though, otherwise we will need to postpone it till 3.0 which is not due for a while.

Check! A patch can be fine too. It depends on how you look at the change, since you could consider this a bug fix :)

yasserf:
awesome, will plan a release to coincide with deepstream.io asap, just pending changelogs

PR #329, opened 2017-02-16
json-path: Optimized patch

ronag:
This fixes a future TODO on jsonPath.patch where previously deepEquals was excessivly called.


yasserf:
Looks legit and all tests seem happy. Thanks!

PR #333, opened 2017-02-19
update component-emitter

ronag:
component/emitter doesn't seem to be actively maintained anymore so I've made a fork.

https://www.npmjs.com/package/component-emitter2

Would like bring attention to the following commits relative to 1.1.2 which deepstream currently uses:

nxtedition/emitter@b34d74c
nxtedition/emitter@7aa9d9b

and

nxtedition/emitter@9baa392
nxtedition/emitter@2d1744a

Also, this fixes the "problem" of deepstream depending on the internal implementation by providing a `eventNames` method.


yasserf:
This is awesome. Thank you for maintaining that and the memory optimizations.

PR #336, opened 2017-02-21
Fix #335: record.set now directly callbacks with no error if old and new value was the same

### LINKED ISSUE #335: record.set not callbacked when data has not changed
**Issue**
The callback added to record.set will only be called back if the client considers the data to be new.

**Reasoning**
As we see it, the user of the API wants to set X and then do something when record has value X. If this results in data not being changed due to old value already being X should not matter, since this is an optimisation.

escwald:
#335


escwald:
I have also added a feature test for this here https://github.com/deepstreamIO/deepstream.io-client-specs/pull/23

escwald:
Added a new PR https://github.com/deepstreamIO/deepstream.io-client-js/pull/339 that fixes this with setImmediate. It also has a fallback to setTimeout(func, 0), since not all browsers support setImmediate.

Closing this PR.

PR #339, opened 2017-02-23
Fix #335: Set immediate callback fix for record.set

### LINKED ISSUE #335: record.set not callbacked when data has not changed

escwald:
A more sophisticated fix for issue #335 instead of PR https://github.com/deepstreamIO/deepstream.io-client-js/pull/336, to use setImmediate instead of synchronously calling callback.

Added a setImmediate in utils that fallbacks to setTimeout(func, 0) if setImmediate is not supported. Also added two unit-tests.

The spec tests at https://github.com/deepstreamIO/deepstream.io-client-specs/pull/23 should still work with this fix.

### DIFF
--- a/src/record/record.js
+++ b/src/record/record.js
@@ -119,7 +119,7 @@ Record.prototype.set = function( pathOrData, dataOrCallback, callback ) {
 		if( ( typeof pathOrData === 'string' && pathOrData.length !== 0 ) && typeof dataOrCallback !== 'function' ) {
 			path = pathOrData;
 			data = dataOrCallback
-		} 
+		}
 		// set( data, callback )
 		else if( typeof pathOrData === 'object' && typeof dataOrCallback === 'function' ) {
 			data = pathOrData;
@@ -151,17 +151,18 @@ Record.prototype.set = function( pathOrData, dataOrCallback, callback ) {
 	var newValue = jsonPath.set( oldValue, path, data, this._options.recordDeepCopy );
 
 	if ( oldValue === newValue ) {
+		if ( typeof callback === 'function' ) utils.setImmediate( callback, null );
 		return this;
 	}
 
 	var config;
-	if( callback !== undefined ) {
+	if( typeof callback === 'function' ) {
 		config = {};
 		config.writeSuccess = true;
 		this._setUpCallback(this.version, callback)
 		var connectionState = this._client.getConnectionState();
 		if( connectionState === C.CONNECTION_STATE.CLOSED || connectionState === C.CONNECTION_STATE.RECONNECTING ) {
-			callback( 'Connection error: error updating record as connection was closed' );
+			utils.setImmediate( callback, new Error( 'Connection error: error updating record as connection was closed' ) );
 		}
 	}
 	this._sendUpdate( path, data, config );
@@ -369,7 +370,7 @@ Record.prototype._recoverRecord = function( remoteVersion, remoteData, message )
 };
 
 Record.prototype._sendUpdate = function ( path, data, config ) {
-	this.version++; 
+	this.version++;
 	var msgData;
 	if( !path ) {
 		msgData = config === undefined ?
@@ -409,7 +410,7 @@ Record.prototype._onRecordRecovered = function( remoteVersion, remoteData, messa
 
 		var config = message.data[ 4 ];
 		if( config && JSON.parse( config ).writeSuccess ) {
-			var callback = this._writeCallbacks[ oldVersion ]; 
+			var callback = this._writeCallbacks[ oldVersion ];
 			delete this._writeCallbacks[ oldVersion ];
 			this._setUpCallback( this.version, callback )
 		}
--- a/src/record/record.js
+++ b/src/record/record.js
@@ -151,7 +151,13 @@ Record.prototype.set = function( pathOrData, dataOrCallback, callback ) {
 	var newValue = jsonPath.set( oldValue, path, data, this._options.recordDeepCopy );
 
 	if ( oldValue === newValue ) {
-		if ( typeof callback === 'function' ) utils.setImmediate( callback, null );
+		if ( typeof callback === 'function' ) {
+			if (isConnected( this._client )) {
+				utils.setImmediate( callback, null );
+			} else {
+				utils.setImmediate( callback, 'Connection error: error updating record as connection was closed' );
+			}
+		}
 		return this;
 	}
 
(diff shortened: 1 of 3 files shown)


ronag:
This will break things in user land. -1

ronag:
LGTM

src/utils/utils.js
```diff
@@ -176,6 +176,22 @@ exports.setInterval = function( callback, intervalDuration ) {
...
+ * This method is used to break up long running operations and run a callback function immediately
+ * after the browser has completed other operations such as events and display updates.
+ *
+ * @param {Function} callback        the function that will be called after the given time
+ * @param {...*}   param1, ..., paramN additional parameters which are passed through to the callback
+ *
+ * @public
+ */
+exports.setImmediate = typeof setImmediate === 'function' ? setImmediate : function() {
+	var args = Array.prototype.slice.call( arguments );
```
yasserf: interesting, the pseudo code is the same as process.nextTick. This is better from a queuing concept though.

src/record/record.js
```diff
@@ -151,17 +151,18 @@ Record.prototype.set = function( pathOrData, dataOrCallback, callback ) {
...
 	var config;
-	if( callback !== undefined ) {
+	if( typeof callback === 'function' ) {
 		config = {};
 		config.writeSuccess = true;
 		this._setUpCallback(this.version, callback)
 		var connectionState = this._client.getConnectionState();
 		if( connectionState === C.CONNECTION_STATE.CLOSED || connectionState === C.CONNECTION_STATE.RECONNECTING ) {
-			callback( 'Connection error: error updating record as connection was closed' );
+			utils.setImmediate( callback, new Error( 'Connection error: error updating record as connection was closed' ) );
```
yasserf: we don't do this in alot of places. TBH I do prefer the concept of having things always get called async rather than  maybe, maybe not.

We don't however encapsulate things inside of error objects,, almost all errors are always from the server and hence are just strings.

src/record/record.js
```diff
@@ -151,17 +151,18 @@ Record.prototype.set = function( pathOrData, dataOrCallback, callback ) {
 	var newValue = jsonPath.set( oldValue, path, data, this._options.recordDeepCopy );
 
 	if ( oldValue === newValue ) {
+		if ( typeof callback === 'function' ) utils.setImmediate( callback, null );
```
yasserf: isn't this a bit odd if your connection goes down, you try and write twice (further down), first time errors, seconds succeeds?

src/record/record.js
ralphtheninja: Maybe we should keep using strings for now so it's consistent. I'd love to see a rewrite in the future where the client uses Error for everything. But in that case the client should probably be released with a new major version.

src/record/record.js
escwald: I don't really understand this comment, can you elaborate? :)

As I see it, it does not result in a new write when the value is the same, since it does a return on next line. The issue was however that it never called the callback in this case.

Resulting in the scenario ```record.set(something, something, function doStuff)``` where _doStuff_ was not always called.

escwald:
I removed the new Error encapsulation, based on previous discussions.

src/record/record.js
yasserf: If the connection goes down, calling set will change the data, and buffer the update change to the server.

If the same data is set again, the line you added will return a fake positive.

yasserf:
```
if( connectionState === C.CONNECTION_STATE.CLOSED || connectionState === C.CONNECTION_STATE.RECONNECTING ) {
 -			callback( 'Connection error: error updating record as connection was closed' );		 +			utils.setImmediate( callback, 'Connection error: error updating record as connection was closed' );
  		}		  		}
```

That check should be done before the check

yasserf: Any thoughts on this edge case? We could just give an error when doing a set when connection is down. Not very offline friendly but thats part of another story anyways.

escwald: Yes, I would rather see that. Since as it is now, the user of the interface will already be notified with an error that the connection is closed and then it tries to set the data anyway. If offline should be supported, it would be a warning at most stating that you are in offline mode.

So I would recommend moving the connectionState check further up in the set function and do a setImmediate on the callback with the error and then return this.

Is this what you meant?

yasserf: Yup, I think that's what we should do. If you do a write and your offline, you get told immediately it failed because it was offline. Doesn't meant the write was aborted though, just that your offline so you won't be told when its done.

escwald:
Updated code to fix the edge-case that was previously discussed.
Also added a test scenario that verifies that it works, see https://github.com/deepstreamIO/deepstream.io-client-specs/pull/23

yasserf:
This looks great! I'll try and get it in asap. I formatted the JS client so it might be a bit painful haha, but this provides a good reason to release!

src/record/record.js
```diff
@@ -151,24 +151,37 @@ Record.prototype.set = function( pathOrData, dataOrCallback, callback ) {
...
 		}
 	}
+
 	this._sendUpdate( path, data, config );
 	this._applyChange( newValue );
 	return this;
 };
 
+function isConnected( client ) {
+	var connectionState = client.getConnectionState();
```
yasserf: Another small question, would we want to callback with an error whenever the connection is not OPEN, as currently your write ack would be delayed until an authentication happens.

escwald: You mean if the connection goes down after line 170? Sounds like a very narrow corner case and can probably be improved later if see fit. Or do you see a very urgent need for this?

yasserf: No I meant if you do a set before login occurs, state is authenticating or awaiting authentication

yasserf: No I meant if you do a set before login occurs, state is authenticating or awaiting authentication

yasserf:
Sorry for the delay getting this in. Just got merge conflicts fixed as well, will be looking to merge this in and release them both asap!

yasserf:
Closing in favor of #352 due to linting

So this works, but I'm a little concerned about this: 

https://developer.mozilla.org/en/docs/Web/API/Window/setImmediate

Thoughts? Happy to merge anyways since it works

yasserf:
So I think I will be using the requestIdleCallback instead of immediate, since we will be introducing that and it seems to be approved as an API. Will refactor this shortly and then merge! Thanks again!

I do apologize if you already did, but it states you are a first-time contributor, would you be okay with signing the CLA?

https://deepstream.typeform.com/to/PwwNd5

escwald:
I have signed the CLA now.

It is okay to use **requestIdleCallback** if it works, but you will probably have to do a fallback method for that as well, since it is also not fully supported.

yasserf:
@escwald yup, theres another PR that heavily relies it so will do

I hope the release time isn't blocking you on anything, we are going to do a minor release with quite a few fixes and performance improvements, but there has been alot of activity in the last week (which is awesome) and trying to get everything in is quite time consuming! This will def make it into the next release which I'm planning early this week though!

yasserf:
Finally just merged this is. I do apologize for the timing, I was off for a short while. 

Anywho, thanks again for raising this, will be doing a release very soon!

PR #342, opened 2017-02-26
Central acktimeout registry

yasserf:


ronag:
Please fix the code style! It's difficult to merge this in with our own code :(.

yasserf:
we will be applying the deepstream code style on this soon, which should make it almost identical

PR #343, opened 2017-03-01
dist: updating distribution files for 2.1.2

AlexBHarley:


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - optimized json-path patch [#329](https://github.com/deepstreamIO/deepstream.io-client-js/pull/329)
 - TypeScript typings [#283](https://github.com/deepstreamIO/deepstream.io-client-js/pull/283) and [#338](https://github.com/deepstreamIO/deepstream.io-client-js/pull/338)
 - Added support for non-NaNish base 16 numbers in jsonpath [#328](https://github.com/deepstreamIO/deepstream.io-client-js/pull/328)
+- There is now a single ack timeout registry for the client, shared between all handlers [#342](https://github.com/deepstreamIO/deepstream.io-client-js/pull/342)
```
yasserf: Ack timeouts do not occur when connection is not open
Ack timeouts get cleared down when connection is lost

yasserf: Just realized a couple of edge cases where doing a snapshot or a has would be wierd when a connection drops because it won't fail due to a timeout.. but snapshots shouldn't anyways

AlexBHarley: @yasserf, comments addressed

PR #348, opened 2017-03-12
Update client.d.ts for Record and List unsubscribe()

EnigmaCurry:
unsubscribe callback is an optional argument.


yasserf:
Nice, thanks!

PR #351, opened 2017-03-20
fix: Invalid remote wins merge conflicts

### LINKED ISSUE #224: merge with local wins is never sent to server
As discussed in #214. Because `set` equality check traps the update.

yasserf:
When the strategy is remote wins and the content is identical, we expect that
the server version number is taken and no actual merge conflict occurs

Fixes #224

### DIFF
--- a/src/record/record.js
+++ b/src/record/record.js
@@ -6,6 +6,7 @@ const EventEmitter = require('component-emitter2')
 const C = require('../constants/constants')
 const messageBuilder = require('../message/message-builder')
 const messageParser = require('../message/message-parser')
+const utils = require('../utils/utils')
 
 /**
  * This class represents a single record - an observable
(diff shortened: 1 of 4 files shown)


src/record/record.js
```diff
@@ -431,7 +432,8 @@ Record.prototype._onRecordRecovered = function (remoteVersion, remoteData, messa
 
     const oldValue = this._$data
     const newValue = jsonPath.set(oldValue, undefined, data, false)
-    if (oldValue === newValue) {
+
+    if (utils.deepEquals(oldValue, remoteData)) {
       return
```
ronag: This shouldn't be required... if `deepEquals` and `oldValue === newValue` differs then there is a bug in `jsonPath.set` and we should have a unit test for that added..

yasserf:
@ronag it said you made a comment but I can't see it anywhere =D

I'm okay with changing the implementation, the pain was getting the tests correct. It become obviously wrong when I noticed I was getting permission issues when my connection dropped and it was trying to tell the server it accepted.. the server data when it didn't have write access.

PR #353, opened 2017-03-25
Client check to invalid update from being sent.

datasage:
This is the client check to prevent an invalid update from being sent and thus saving a round trip.

Related PR: https://github.com/deepstreamIO/deepstream.io/pull/608

### DIFF
--- a/src/record/record.js
+++ b/src/record/record.js
@@ -150,6 +150,10 @@ Record.prototype.set = function (pathOrData, dataOrCallback, callback) {
     data = dataOrCallback
   }
 
+  if(!path && (data === null || typeof data !== 'object')) {
+      throw new Error('invalid arguments, scalar values cannot be set without path');
+  }
+
   if (this._checkDestroyed('set')) {
     return this
   }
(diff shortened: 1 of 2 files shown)


yasserf:
I believe this might break Lists

https://github.com/datasage/deepstream.io-client-js/blob/67a103f41224c9a32046498ccad4e525f579949b/src/record/list.js#L101

Where it needs to accept an array type

datasage:
A typeof check on array returns 'object' so it will work in this case. It would also allow through `new String('my string')`, which would then get blocked by the server. 

There are a couple ways to check for that, but they have their downsides and may not be worth it on the client.

Could block these by checking the constructor, but that could get into prevent valid uses where where someone is using a class with a constructor other than Object or Array, that does encode to a JSON object.

The other alternative is to encode/parse the value as a JSON, and check the data type of the result. But that would have a significant performance impact depending on the size of the object.

The server should be the primary check for this anyway. This just provides more immediate feedback for some of the most obvious cases.

yasserf:
Good point, totally forgot that for a second. I'm happy to merge this, pending the CLA mentioned on the other PR

yasserf:
Merging, thanks!

PR #355, opened 2017-03-26
Bad login message causes reconnect loop

datasage:
So if you send a bad login message, you get an INVALID_AUTH_MSG and the connection with the server closes. This was not being treated as a final state and it was attempting to reconnect indefinitely.

I added handling for INVALID_AUTH_MSG to mark the connection as closed deliberately and some best effort detection to ensure authParams is actually an object.


yasserf:
Hey, good point. Closing connections with invalid message data during the connection and authentication state makes sense.

This PR looks good, merging in!

test-unit/unit/message/connectionSpec.js
```diff
@@ -353,6 +353,51 @@ describe('connection handles auth rejections', () => {
...
+    expect(connection.getState()).toBe('AWAITING_CONNECTION')
+  })
+
+  it('sends auth parameters on connection ack', () => {
+    authCallback = jasmine.createSpy('invalid auth callback')
+
+    connection._endpoint.emit('message', msg('C|A+'))
+    connection.authenticate('Bad Auth', authCallback)
+
+    expect(clientMock.lastError).toEqual(['X', 'INVALID_AUTH_MSG', 'authParams is not an object'])
```
yasserf: could we add a check here that nothing actually got sent to the server?

yasserf: expect(connection._endpoint.lastSendMessage).toBe(null)

yasserf:
Found one little extra line which would be awesome in tests

test-unit/unit/message/connectionSpec.js
datasage: Yeah, give me a few minutes to add this.

datasage: Actually the previous test does set a message, is there a way to rest/clear that for the next text?

yasserf: Should be able to either use a `beforeEach` at the top of the describe or just null it at the beginning of the it test

yasserf:
Looks good, will merge once green. Thanks!

PR #359, opened 2017-03-29
Typescript additions

EnigmaCurry:
Here's a few more changes to the typescript defs. I think we're getting pretty close to the docs now :)


yasserf:
cool, thanks!

PR #361, opened 2017-04-02
Request idle callback

yasserf:
@ronag can't get a few small tests to pass =(


yasserf:
This annoying has just a couple of tests left to fix until I can merge in.

If anyone would like to give it a shot would be awesome. It should make a huge difference in browser lag with extreme load once its in.

PR #368, opened 2017-04-19
Missing EVENT in tryReconnect and change in connection state after max reconnections attempt reached

rbarroetavena:
fix(connection): add missing EVENT to _tryReconnect MAX_RECONNECTION_ATTEMPTS_REACHED and unit test

fix(connection): leave connection closed after MAX_RECONNECTION_ATTEMPTS_REACHED

Thanks


jdmnd:
Hey @rbarroetavena. Thanks for the contribution!

Would you mind perhaps signing out [Contributor License Agreement](https://deepstream.io/info/community/cla/)?

Thanks!
John

src/message/connection.js
```diff
@@ -531,7 +531,8 @@ Connection.prototype._tryReconnect = function () {
   } else {
     this._clearReconnect()
     this.close()
-    this._client.emit(C.MAX_RECONNECTION_ATTEMPTS_REACHED, this._reconnectionAttempt)
+    this._setState(C.CONNECTION_STATE.CLOSED)
```
jdmnd: Was there an issue that motivated the call to `_setState`? 
The state is also updated asynchronously in the call to `_onClose` (once the endpoint closes).

rbarroetavena: Hi @jdmnd 
I've accepted the license.
regarding the state change we're using JWT token authentication. We always need to ensure fresh tokens (lifetime 1min or 5min) during auth.

First we were not using the embedded reconnect mechanism: _options.maxReconnectAttempts = 1. Reconnection was handled from outside the ds-io.client, so we needed the connection to end in state CLOSED instead of ERROR. Connection CLOSED allowed us to call login() on the same deepstream instance with a fresh token.  That's why we added the _setState(C.CONNECTION_STATE.CLOSED) when MAX_RECONNECTION_ATTEMPTS_REACHED.

Now we've changed our implementation and started using the embedded reconnection mechanism (retry forever). Right after connection changes to AWAITING_AUTHENTICATION, we set _connection._authParams to null. This gives us the time we need to go get a fresh token and then call login.

We're not using the _setState anymore. If you think it's better to leave the connection in the ERROR state pls let me know and i'll update the pr. 

Thanks and sorry for the lengthy comment.

jdmnd: I'd like to get this into a release I'm doing today so if you'd could remove the setState call that would be great.
From looking at the code I can't see what would put the connection state into ERROR when maxReconnectionAttempts is reached so that sounds like a bug. Would you be able to put instructions to replicate into an issue?
Thanks!

rbarroetavena: i've just removed the setState.
i'll check again final MAX_RECONNECTION_ATTEMPTS_REACHED connection state and let you know. 
thanks

PR #372, opened 2017-05-02
Correctly handle `login(callback)`

### LINKED ISSUE #371: Login with callback and no auth data sends callback to server as auth data
```js
deepstream('localhost:6020')
  .login(() => console.log('logged in'))
```
results in
```
Uncaught Error: INVALID_AUTH_MSG: authParams is not an object (X)
    at Client._$onError (deepstream.js:2040)
    at Connection.authenticate (deepstream.js:2696)
    at Client.login (deepstream.js:1920)
    at <anonymous>:1:35
```
Expected: 
The callback should be called on auth response

jdmnd:
Fixes #371


yasserf:
Nice, you would also need to add the typescript definitions (both arguments are options) and if possible a single UT?

PR #386, opened 2017-06-08
Try to stringify message errors so socket errors keep their data.

SejH:
Socket errors are returned to the client as the string "[object Event]".


jdmnd:
Hey @SejH 
Thanks for the contribution!

1. Would you be able to add a unit test that demonstrates the issue?
2. The `try/catch` block here is not required, since `JSON.stringify` never throws exceptions.

yasserf:
Would you be okay signing our CLA? I know its a pretty small change but its easier for us to get all PRs in. Thank you!

https://deepstream.typeform.com/to/PwwNd5

yasserf:
Small reminder, be awesome if you could sign the CLA, we can't merge anything in without it =(

https://deepstream.typeform.com/to/PwwNd5

jdmnd:
CLA signature received, thanks!

PR #403, opened 2017-09-26
Promisified API

AlexBHarley:


yasserf:
This looks awesome! Happy to merge

AlexBHarley:
Promisified APIs are untested. Bit difficult with how we use callbacks in tests as well. Implementation is very simple but should probably not be released for a little bit.

PR #410, opened 2017-10-26
Feat: List methods with write acknowledgement

svagi:
- extends `setEntries`, `addEntry` & `removeEntry` methods with callback argument which is passed to `record.set`
- adds new promisifed methods: `setEntriesWithAck`, `addEntryWithAck` & `removeEntryWithAck`

See: https://github.com/deepstreamIO/deepstream.io/issues/830

### DIFF
--- a/src/record/list.js
+++ b/src/record/list.js
@@ -78,10 +78,13 @@ List.prototype.isEmpty = function () {
 /**
  * Updates the list with a new set of entries
  *
- * @public
  * @param {Array} entries
+ * @param {Function} [callback]
+ *
+ * @public
+ * @returns {void}
  */
-List.prototype.setEntries = function (entries) {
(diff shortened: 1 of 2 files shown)


AlexBHarley:
Like I said in [#830](https://github.com/deepstreamIO/deepstream.io/issues/830). Can we just return a promise when no callback is provided?

Ideally we wouldn't add three more methods to the API.

svagi:
@yasserf @AlexBHarley What do you think?

yasserf:
Generally happy with this. @AlexBHarley? Not sure if we want to do a release b4 v4 with this but makes alot of sense to merge this in to make sure we get it into v4

yasserf:
@svagi thanks for raising this!

PR #413, opened 2017-11-02
V4 test listener

MuriloFrade:


test/mocks.ts
```diff
@@ -84,3 +84,16 @@ export const getServicesMock = () => {
     }
   }
 }
+
+export const getListenerMock = () => {
+  const listener = {
```
yasserf: you can do this instead (i think)

listenerMock = mock(listener.prototype)

means it will always use the correct API. But it doesn't change anything otherwise

PR #418, opened 2017-11-03
V4 rpc tests

MuriloFrade:


src/rpc/rpc-response.ts
```diff
@@ -16,7 +16,8 @@ export class RPCResponse {
     private correlationId: string
     private isAccepted: boolean
     private isComplete: boolean
-    private autoAccept: boolean
+
+    public autoAccept: boolean
```
yasserf: why is this public?

test/rpc/rpc-handlerSpec.ts
```diff
@@ -0,0 +1,344 @@
...
+      topic: TOPIC.RPC,
+      action: RPC_ACTIONS.REQUEST,
+      name,
+      parsedData: data,
+      correlationId: '123'
+    })
+  })
+
+  it('doesn\'t send messages and warn on auth/denied errors', () => {
+    services.connectionMock
```
yasserf: this doesn't make any sense

test/rpc/rpc-handlerSpec.ts
```diff
@@ -0,0 +1,344 @@
...
+        parsedData: data
+      })
+
+      await BBPromise.delay(rpcResponseTimeout * 2)
+      sinon.assert.calledOnce(rpcResponse)
+      sinon.assert.calledWithExactly(rpcResponse, null, data)
+    })
+
+    it('doesn\'t call rpc response callback twice when handling response message', async () => {
+      handle({
```
yasserf: message = {
}
handle(message)
handle(message)

yasserf:
need to test message denied or error, need to test promises

test/rpc/rpc-handlerSpec.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    it('doesn\'t register provider twice', () => {
+      services.connectionMock
+        .expects('sendMessage')
+        .never()
+      services.timeoutRegistryMock
+        .expects('add')
+        .never()
+
+      expect(rpcHandler.provide.bind(rpcHandler, name, rpcProviderSpy as RPCProvider)).to.throw()
```
yasserf: what does this throw?

test/rpc/rpc-handlerSpec.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    it('replies requests', () => {
+      const message: RPCMessage = {
+        topic: TOPIC.RPC,
+        action: RPC_ACTIONS.REQUEST,
+        name,
+        parsedData: data,
+        correlationId: '123'
+      }
+      const rpcResponse = new RPCResponse(message, options, services)
```
yasserf: this makes little sense

test/rpc/rpc-responseSpec.ts
```diff
@@ -0,0 +1,138 @@
...
+    )
+    rpcResponse.autoAccept = false
+  })
+
+  afterEach(() => {
+    services.connectionMock.verify()
+  })
+
+  it('doesn\'t accept automatically when autoAccept == false', () => {
+    services.connectionMock
```
yasserf: this does not make sense

yasserf: unless the ack is async, in which case it does, but then the test needs to be async or it will obviously never be called

test/rpc/rpc-handlerSpec.ts
```diff
@@ -0,0 +1,344 @@
...
+        name,
+        parsedData: data,
+        correlationId: sinon.match.any
+    })
+
+    rpcHandler.make(name, data, () => {})
+  })
+
+  it('returns promise on make when no callback is passed', () => {
+    services.connectionMock
```
yasserf: should test the promise itself.. not just see it returns one

test/rpc/rpc-responseSpec.ts
```diff
@@ -0,0 +1,140 @@
...
+      .withExactArgs(message)
+
+    rpcResponse.send(data)
+  })
+
+  it('throws when trying to send a completed response', () => {
+    const data = { foo: 'bar'}
+    services.connectionMock
+      .expects('sendMessage')
+      .twice()
```
yasserf: why would this send two messages?

test/rpc/rpc-responseSpec.ts
```diff
@@ -0,0 +1,140 @@
...
+        topic: TOPIC.RPC,
+        action: RPC_ACTIONS.REJECT,
+        name,
+        correlationId
+      })
+
+    rpcResponse.reject()
+  })
+
+  it('sends error message', () => {
```
yasserf: rejecting twice

test/rpc/rpc-responseSpec.ts
```diff
@@ -0,0 +1,140 @@
...
+    services.connectionMock
+      .expects('sendMessage')
+      .once()
+
+    rpcResponse.accept()
+    rpcResponse.accept()
+  })
+
+  it('sends reject message', () => {
+    services.connectionMock
```
yasserf: rejecting twice

test/rpc/rpc-responseSpec.ts
yasserf: if one is for auto accept than we should put it in a comment i guess

src/rpc/rpc-handler.ts
```diff
@@ -90,7 +90,7 @@ export class RPCHandler {
    * @param   {Function} callback Will be invoked with the returned result or if the rpc failed
    *                              receives to arguments: error or null and the result
    */
-  public make (name: string, data: any, callback: (error: string, data: any) => void): Promise<any> | undefined {
+  public make (name: string, data: any, callback?: (error: string, data: any) => void): Promise<any> | undefined {
     if (typeof name !== 'string' || name.length === 0) {
```
yasserf: if (this.services.connection.isConnected) {
  if (callback) {
    callback(EVENT.CLIENT_OFFLINE)
  } else {
    return promise that rejects
  }
}

yasserf:
looks good other than a couple small things. Nice work!

src/rpc/rpc-handler.ts
MuriloFrade: What about the provide and unprovide calls when offline ?

test/rpc/rpc-responseSpec.ts
MuriloFrade: I've made this tests more clear for the reader now

yasserf:
Failing tests

src/rpc/rpc-handler.ts
```diff
@@ -90,11 +90,22 @@ export class RPCHandler {
    * @param   {Function} callback Will be invoked with the returned result or if the rpc failed
    *                              receives to arguments: error or null and the result
    */
-  public make (name: string, data: any, callback: (error: string, data: any) => void): Promise<any> | undefined {
+  public make (name: string, data: any, callback?: RPCMakeCallback): Promise<any> | undefined {
     if (typeof name !== 'string' || name.length === 0) {
       throw new Error('invalid argument name')
     }
 
+    const cb = callback && typeof callback === 'function'
```
yasserf: If callback is provided and is not a function should error, not fail silently

test/rpc/rpc-handlerSpec.ts
yasserf: This test hasn’t changed yet.

test/rpc/rpc-responseSpec.ts
yasserf: Did that just work with once and twice? If so something is wrong with test verification.

test/rpc/rpc-handlerSpec.ts
MuriloFrade: The promises for all make requests are tested in the 'when making' describe block. I think it doesn't make sense to keep this since all the tests in 'when making' will fail if the returning object is not a promise.

test/rpc/rpc-responseSpec.ts
MuriloFrade: Sorry Yasser, but in which test do you mean once and twice now?

yasserf: The test initially had .twice and now its .once

Why did it work with .twice before?

src/rpc/rpc-handler.ts
```diff
@@ -90,10 +90,21 @@ export class RPCHandler {
...
     if (typeof name !== 'string' || name.length === 0) {
       throw new Error('invalid argument name')
     }
+    if (callback && typeof callback !== 'function') {
+      throw new Error('invalid argument callback')
+    }
+
+    if (this.services.connection.isConnected === false) {
+      if (callback) {
+        callback(EVENT.CLIENT_OFFLINE)
```
yasserf: timerService.requestNextIdle...

src/rpc/rpc-handler.ts
```diff
@@ -90,10 +90,21 @@ export class RPCHandler {
...
+  public make (name: string, data: any, callback?: RPCMakeCallback): Promise<any> | undefined {
     if (typeof name !== 'string' || name.length === 0) {
       throw new Error('invalid argument name')
     }
+    if (callback && typeof callback !== 'function') {
+      throw new Error('invalid argument callback')
+    }
+
+    if (this.services.connection.isConnected === false) {
+      if (callback) {
```
yasserf: timerService.requestNextIdle...

src/rpc/rpc-handler.ts
```diff
@@ -103,7 +103,9 @@ export class RPCHandler {
         callback(EVENT.CLIENT_OFFLINE)
```
jdmnd: doesn't this need to be async too?

src/rpc/rpc-handler.ts
```diff
@@ -103,7 +103,9 @@ export class RPCHandler {
...
         return
       }
-      return Promise.reject(EVENT.CLIENT_OFFLINE)
+      return new Promise((resolve, reject) => {
+        nextTick(reject.bind(this, EVENT.CLIENT_OFFLINE))
+      })
```
jdmnd: also i think promises don't resolve synchronously, so this is unnecessary
https://jsblog.insiderattack.net/promises-next-ticks-and-immediates-nodejs-event-loop-part-3-9226cbe7a6aa

PR #419, opened 2017-11-07
V4 utils

jshrt:
* refactored normalizeSetArguments to accept a data argument to be all types except function or undefined 
* added normalizeSetArguments test

### DIFF
--- a/src/util/utils.ts
+++ b/src/util/utils.ts
@@ -181,65 +181,69 @@ export const getUid = (): string => {
 
 export interface RecordSetArguments { callback?: (error: string | null) => void, path?: string, data?: any }
 export interface RecordSubscribeArguments { callback: (data: any) => void, path?: string, triggerNow?: boolean }
-  /**
-   * Creates a map based on the types of the provided arguments
-   */
+
+/**
+ * Creates a map based on the types of the provided arguments
+ */
 export const normalizeSetArguments = (args: IArguments, startIndex: number = 0): RecordSetArguments => {
-    let result
+  let result
 
-    if (args.length === startIndex + 1) {
-      result = { data: args[startIndex], path: undefined, callback: undefined }
-    }
+  const isData = (data: any) => data !== undefined && typeof data !== 'function'
+  const isPath = (path: any) => path !== undefined && typeof path === 'string'
+  const isCallback = (callback: any) => typeof callback === 'function'
 
-    if (args.length === startIndex + 2) {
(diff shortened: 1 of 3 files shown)


src/util/utils.ts
```diff
@@ -181,65 +181,87 @@ export const getUid = (): string => {
...
-      }
-      if (result.callback !== undefined && typeof result.callback !== 'function') {
-        throw Error ('Invalid set callback argument')
-      }
-      if (result.data === undefined) {
-        throw Error ('Invalid set data argument')
-      }
-      return result
+    if (!isPath(args[startIndex])) {
+      result.callback = isCallback(args[startIndex + 1]) ? args[startIndex + 1] : false
```
yasserf: Why is callback a function or false?

src/util/utils.ts
```diff
@@ -181,65 +181,87 @@ export const getUid = (): string => {
...
     }
+  }
 
-    throw Error ('Invalid set arguments')
+  if (args.length === startIndex + 3) {
+    result = {
+      path: isPath(args[startIndex]) ? args[startIndex] : false,
```
yasserf: How is path false?

jshrt: those `false` values in this case are a way to trigger precise error messages, as `undefined` can be an acceptable case for args.length cases

jshrt:
The refactor takes into consideration two use cases:
1. When data is supposed to be at root level of the record, which then only allows it to be of type `object`
2. When data is at a certain path in the record, which then allows it to be of type `object` and also primitives. (_as discussed last time_)

nested method calls feels like too much, but seemed more readable than redundant type checking.

Let me know what to keep and to discard

test/util/utilsSpec.ts
```diff
@@ -576,4 +622,9 @@ describe('parseUrl adds all missing parts of the url', () => {
     expect(utils.parseUrl('localhost?query=value#login', '/deepstream'))
       .to.equal('ws://localhost/deepstream?query=value#login')
   })
+
+  it('rejects urls with @no-host', () => {
```
yasserf: I would stop using @ notations in tests, never seen that standard anwhere before.

src/util/utils.ts
yasserf: Cool, makes sense now

test/util/utilsSpec.ts
jshrt: I agree.

Those are basically meaningless leftovers that I forgot to remove, I was using them with `--grep` when I was focusing on these cases..

PR #426, opened 2017-11-10
Webpack: frontend bundle sockets, deepstream named export

jshrt:
It seems that when we try to use `WebSocket` as our frontend `ws` library the following error happens:

screenshot from 2017-11-10 11-56-46

We should use a proper default socket library for frontend in case whoever is using our bundle does not feel like using his own socket factory

This pr should be merged when the frontend bundle is 100% ready


jshrt:
Closed in favor of https://github.com/deepstreamIO/deepstream.io-client-js/pull/435

PR #429, opened 2017-11-10
V4 record: has, head and snapshot

MuriloFrade:


test/record/list-onlineSpec.ts
```diff
@@ -0,0 +1,234 @@
+// import { Promise as BBPromise } from 'bluebird'
```
yasserf: you can just do describe.skip btw.

yasserf:
looks good, although maybe squashing promise and callback tests together is a little more confusing in terms of assertions per test.

PR #431, opened 2017-11-14
test: fixed connection lost test

jshrt:


test/util/timeout-registrySpec.ts
```diff
@@ -121,7 +126,10 @@ describe('timeout registry', () => {
 
         it('clears timeout when connection lost', done => {
```
MuriloFrade: Does this test do the same thing that the test above? They have the same description.

test/util/timeout-registrySpec.ts
```diff
@@ -121,7 +126,10 @@ describe('timeout registry', () => {
...
             services.simulateConnectionLost()
-            setTimeout(done, 10)
+            setTimeout(() => {
+                sinon.assert.callCount(spy, 0)
+                done()
+            }, 50)
```
MuriloFrade: What 50 means here ? 

Does it work like that also? 
```
services.simulateConnectionLost()
sinon.assert.notCalled(spy)
timeout(done, delay)
```

test/util/timeout-registrySpec.ts
```diff
@@ -70,6 +71,10 @@ describe('timeout registry', () => {
         })
 
         it('clears timeout when connection lost', done => {
+            services.loggerMock
+                .expects('warn')
+                .never()
+
```
MuriloFrade: You still need to test if the timeoutRegistry has called the timeRegistry.remove service for each timeout, no?

test/connection/connectionSpec.ts
```diff
@@ -289,6 +289,7 @@ describe('connection', () => {
     await sendChallengeResponse()
     await receiveChallengeAccept()
     await receiveConnectionError()
+    await BBPromise.delay(10)
```
MuriloFrade: you could use some variable to store the 10 value (`reconnectIntervalIncrement` for example) and use this in more places. This should make it easy to understand what '10'  means in this test.

test/util/timeout-registrySpec.ts
jshrt: I agree, it should actually assert for that, however for now `onConnectionLost` clears timeouts using `clearTimeout` without relying on the `timerRegistry` service. I wonder whether `onConnectionLost` should be refactored to use the services instead?

yasserf:
status of PR?

PR #459, opened 2018-02-01
fix initialization of settings object

Thomas-P:
I fixed the initialization of the settings object to be typescript conform without any workaround hacks.  As you see, your depends on two services befor beeing fully initialized. I think this should be refactored in the future to avoid side effects, when changing other parts in the code.


yasserf:
Thanks for raising! I put in a similar fix using a Partial since we cast it either ways for both consuming services

PR #460, opened 2018-02-01
Provide a better interface for creation

Thomas-P:
This is an approach for providing a better factory for client generation. This allows TypeScript to recognize the static properties and build the client via new client. In principal, this could be also done in the main client class.


yasserf:
This completely breaks backwards compatibility though..

yasserf:
Really sorry I missed this, the fix is perfect, copying it over due to merge conflict.

yasserf:
Theres actually a little bit of a backwards compatibility issue =(. You seem like a Typescript Guru, how would I manage to keep client creation backwards compatible while also having decent typescript bindings?

PR #461, opened 2018-02-01
fix for connection state change

Thomas-P:
The connection state for close event is not emmited on close. Thats why I added this for a fix in version 2.3.0. Someone has to do this also for 4.0. I think I will commit another one for this.

I could not build it, so someone has to do it, before publishing an new version to npm. There is a problem with the sub branch of urps. I did not have the right to access them.

### DIFF
--- a/src/message/connection.js
+++ b/src/message/connection.js
@@ -144,6 +144,7 @@ Connection.prototype.close = function () {
   clearInterval(this._heartbeatInterval)
   this._deliberateClose = true
   this._endpoint.close()
+  this._setState(C.CONNECTION_STATE.CLOSED)
 }
 
 /**


yasserf:
hey, thanks for raising! Could you raise this vs master instead of v4?

Thomas-P:
Hi, I checked it in Version 4.0 and there it seems to be fixed. 

This is the position in the sources.
[connection.ts -> close()](https://github.com/deepstreamIO/deepstream.io-client-js/blob/v4/src/connection/connection.ts#L267)

yasserf:
Awesome! In thats case u can just close this PRand open it against master (V4 is the default branch for PRs to open again currently)

Thomas-P:
I will close this for #462.

PR #472, opened 2018-10-03
Add discard function to List

CatEars:
Seems like `List` is missing a discard function in DS4. As far as I could see in DS3, lists are basically wrappers for records? As far as I could tell DS3 only set an internal flag (which is no longer there) apart from discarding the record.


yasserf:
Will be merging in your V4 branch shortly and will inherit this! Thanks!

PR #512, opened 2019-10-29
fix: missing resubscription transition state

mhaagens:
Fixes a missing transition state on resubscription which caused errors.


yasserf:
Awesome thanks!

PR #514, opened 2019-11-15
fix: DeepstreamClient.login() overload signature is incorrect

Vortex375:
The overload signature for `login()` was incorrect, causing the following statement to not work as expected:
```
const client = new DeepstreamClient('localhost:6020')
await client.login({ username: 'foo' })
```
Second line yields
```
'await' has no effect on the type of this expression.ts(80007)
```
Because TypeScript resolves the overload that returns `void`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
+## [5.0.5] - 2019.11.21
 
+### Fix
+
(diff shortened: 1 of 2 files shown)


yasserf:
Awesome thanks! Can you also bump up the change log and I’ll just publish the new release!

Vortex375:
Sorry for the delay. I updated the change log.

Issue #515, opened 2019-12-23
React Native 0.59+ URLSearchParams 'Error: not implemented'


adrhub:
Hi, i am trying to upgrade deepstream js client to v4, on React Native 0.61 but application seems to fail on this line of code:
` const client = new DeepstreamClient(url)`

because on `pareseUrl` it is calling `URL.host` and reaches the following code that throws an exception: 
https://github.com/facebook/react-native/blob/e6057095adfdc77ccbbff1c97b1e86b06dae340b/Libraries/Blob/URL.js#L157-L159

![Capture](https://user-images.githubusercontent.com/10502415/71413175-a2977080-2659-11ea-97ac-cb5b07679951.PNG)

Also i get this warning when removing the validation buy it might be another issue not related to this one
![Capture](https://user-images.githubusercontent.com/10502415/71414813-66680e00-2661-11ea-9ad7-6f9a88d02e82.PNG)

I've  tried solutions described here [[0.59.x] URLSearchParams 'Error: not implemented'](https://github.com/facebook/react-native/issues/23922) but nothing seems to work.
Note: Client version 2.3.4 seems to work fine.

Is there any chance to support deepstream client **v4**+ on React Native **v0.59**+ ?
My dependencies are:
```
"react": "16.9.0",
"react-native": "0.61.2",
"@deepstream/client": "^4.1.3",
"url": "^0.11.0"
```

PR #516, opened 2019-12-24
Fix react native parsing url

adrhub:
Please see related issue https://github.com/deepstreamIO/deepstream.io-client-js/issues/515


adrhub:
Please see related issue https://github.com/deepstreamIO/deepstream.io-client-js/issues/515

PR #517, opened 2019-12-24
fix: parsing url and jsonTransportMode warning 

adrhub:
Related issue: https://github.com/deepstreamIO/deepstream.io-client-js/issues/515

### DIFF
--- a/src/connection/socket-factory.ts
+++ b/src/connection/socket-factory.ts
@@ -7,7 +7,7 @@ const BrowserWebsocket = (global.WebSocket || global.MozWebSocket) as any
 
 export type SocketFactory = (url: string, options: JSONObject, heartBeatInterval: number) => Socket
 
-export const socketFactory: SocketFactory = (url, options = { jsonTransportMode: false }, heartBeatInterval) => {
+export const socketFactory: SocketFactory = (url, options = {}, heartBeatInterval) => {
     const socket = BrowserWebsocket
         ? new BrowserWebsocket(url, [], options)
         : new (require('ws'))(url, options) as any
--- a/src/util/utils.ts
+++ b/src/util/utils.ts
@@ -1,10 +1,4 @@
-let UniformResourceLocator: any
-
-if (typeof URL === 'undefined') {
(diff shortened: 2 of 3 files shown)


src/util/utils.ts
```diff
@@ -1,10 +1,4 @@
-let UniformResourceLocator: any
-
-if (typeof URL === 'undefined') {
-  UniformResourceLocator = require('url').URL
-} else {
-  UniformResourceLocator = URL
-}
+const URL: any = require('url')
```
yasserf: I think the idea behind this was to keep the bundle size small by not importing the URL module via webpack

src/connection/socket-factory.ts
```diff
@@ -7,7 +7,7 @@ const BrowserWebsocket = (global.WebSocket || global.MozWebSocket) as any
 
 export type SocketFactory = (url: string, options: JSONObject, heartBeatInterval: number) => Socket
 
-export const socketFactory: SocketFactory = (url, options = { jsonTransportMode: false }, heartBeatInterval) => {
```
yasserf: why was this removed?

yasserf: can we delete it instead from options? If someone sets it to true it would give the same error

src/util/utils.ts
adrhub: I understand but react native doesn't have implementation for URL:
https://github.com/facebook/react-native/blob/e6057095adfdc77ccbbff1c97b1e86b06dae340b/Libraries/Blob/URL.js#L157-L159
It makes it impossible to use this library after this update

src/connection/socket-factory.ts
adrhub: I didn't remove it to support backward compatibility in case someone uses it for some reason, do you think we can remove it ?

src/util/utils.ts
yasserf: Does react native not have the window.URL same as browsers?

adrhub: Unfortunately no

PR #525, opened 2020-05-01
Fix connection timeouts if server doesn't send something in time

jaime-ez:
This prevents the client from hanging when a user tries to write to a record with ack but he only has read access.

### DIFF
--- a/src/connection/connection.spec.ts
+++ b/src/connection/connection.spec.ts
@@ -67,10 +67,10 @@ describe('connection', () => {
     await recieveConnectionClose()
   })
 
-  it('send pong when ping received across all states', async () => {
+  it('send ping and recieve pong across all states', async () => {
     await openConnection()
-    sendPong()
-    receivePing()
+    sendPing()
+    receivePong()
   })
 
   it('miss heartbeat once', async () => {
@@ -504,20 +504,20 @@ describe('connection', () => {
     await PromiseDelay(0)
   }
 
-  function receivePing () {
+  function receivePong () {
     socket.simulateMessages([{
       topic: TOPIC.CONNECTION,
-      action: CONNECTION_ACTION.PING
+      action: CONNECTION_ACTION.PONG
     }])
   }
 
-  function sendPong () {
+  function sendPing () {
     socketMock
       .expects('sendParsedMessage')
       .once()
       .withExactArgs({
         topic: TOPIC.CONNECTION,
-        action: CONNECTION_ACTION.PONG
+        action: CONNECTION_ACTION.PING
       })
   }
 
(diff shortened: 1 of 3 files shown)


jaime-ez:
There is still an issue with the emitter on L.562, is not being called properly when the record.set call is without callback (given the read and not write permission)

yasserf:
Thank you, I’m currently offline (off tech offline) for a week so will get back to this then!

yasserf:
@jaime-ez if you could write a basic client example that shows the issue I can fix it up. Unless it isn’t something easily reproducible, then let me know steps and I’ll put in a test to reproduce it that way (next week)

jaime-ez:
No problem, it's simple to reproduce:

On server set these permissions:
```
presence:
  "*":
    allow: true
record:
  "*":
    create: true
    write: user.id === 'ok'
    read: true
    delete: true
    listen: true
    notify: true
event:
  "*":
    publish: true
    subscribe: true
    listen: true
rpc:
  "*":
    provide: true
    request: true

```

Then execute this on a file, expected behaviour described on code:
```
'use strict'

const { DeepstreamClient } = require('@deepstream/client')

const readWriteClient = new DeepstreamClient('localhost:6020/deepstream')
const readClient = new DeepstreamClient('localhost:6020/deepstream')

// this client can read and write
readWriteClient.login({ username: 'ok' })

// this client can only read
readClient.login()

readWriteClient.on('connectionStateChanged', (s) => {
  if (s === 'OPEN') {
    // we use the client that has write and read permission
    readWriteClient.record.getRecord('test').whenReady((rec) => {
      // everything works as intended
      console.log('record data', rec.get())

      rec.set('hello', 'world', (e) => {
        console.log('no error on set', e === null)
        console.log('updated record', rec.get())

        // here is the issue:
        // now with the client that does not have write permission, only read
        readClient.record.getRecord('test').whenReady((r) => {
          // we can read as intended
          console.log('can read', r.get())

          // we can not write, but an error should be called or emitted, instead the client just hangs
          r.set('bye', 'world', (err) => {
            // THIS SHOULD BE CALLED
            console.log('error is never called', err)
          })
        })
      })
    })
  }
})

readClient.on('error', (e) => {
  // OR AN ERROR EMITED HERE
  console.log('no error emitted', e)
})
```

yasserf:
Amazing thank you, will look into this in a few days!i

yasserf:
lgtm, would merge in except its blocked on 

> There is still an issue with the emitter on L.562, is not being called properly when the record.set call is without callback (given the read and not write permission)

we could break that out into a separate issue and I can this merged and released, not sure if I'm going to be able to look into it right now due

jaime-ez:
Ok for breaking it into a separate issue. Let me know if you need any help with it, last time I wasnt able to solve it.

yasserf:
Will do thanks!

yasserf:
Incredible thank you

PR #530, opened 2020-06-26
[fix] include url package as external dependency

jaime-ez:
Hi @yasserf, hope everything is ok.

This is actually needed for the client to work in react native. I basically recovered the code from [this pull request](https://github.com/deepstreamIO/deepstream.io-client-js/pull/517), added the missing package.json dependency, and ommited the change to `jsonTransportMode` since it just triggers a warning in react native.

I compared the bundle sizes:
with the url package: non-minified 550707 B / minified 189400 B
As is : non-minified 550703 B / minified 189375 B

The difference is negligeable in my opinion.

Let me know what you think and I can push the package and changelog changes.

Cheers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5277,8 +5277,7 @@
     "querystring": {
       "version": "0.2.0",
       "resolved": "https://registry.npmjs.org/querystring/-/querystring-0.2.0.tgz",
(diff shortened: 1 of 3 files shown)


jaime-ez:
I realized there are some webpack issues, will look into them and update this pull in the upcoming days...

Issue #528, opened 2020-05-29
dist directory/.js-file for browser is missing


davidbehler:
I'm trying to upgrade from deepstream 2.4 to 5.x and can't find the appropriate client .js file to include. For 2.x the releases included a dist directory with the compiled .js files that could be used in the browser, but for the 5.x releases this folder and the compiled .js files are missing.

Is there a reason for this? If so, what's the best way for me to get the client .js I need?

davidbehler:
I found this URL in on of your [tutorials](https://deepstream.io/tutorials/getting-started/javascript/): https://cdn.deepstream.io/js/client/latest/ds.min.js

Is that the right way to go?

jaime-ez:
You can install from npm with `npm install @deepstream/client`, or clone the repo and execute `npm install && npm run tsc` and the dist folder will be compiled

davidbehler:
I did that, but neither deepstream.js nor deepstream-client.js expose a DeepstreamClient constructor when including via script-tag into my website.

They both look significantly different than the file on the cdn and I think I'm missing a processing step or something.

jaime-ez:
I ran this index.html file: 
```
<!DOCTYPE html>
<html lang="en" dir="ltr">
  <head>
    <meta charset="utf-8">
    <title></title>
    <script src="https://cdn.deepstream.io/js/client/latest/ds.min.js" charset="utf-8">
    </script>
  </head>
  <body>
    <p>HI</p>
    <script type="text/javascript">
      var client = new window.DeepstreamClient('localhost:6020')
      client.login({}, function(success) {
        console.log('success', success)
      })
    </script>
  </body>
</html>
```

While runngin an open deepstream server in port 6020 and had no issues, the client connected succesfully.

davidbehler:
Yes, the file from CDN works fine. But no where is it explicitly stated that the file from CDN should be used. 

Previous (pre 5.x) releases included a ready-to-use .js file and that's what I was looking for, since we are upgrading to 5.x.

This issue can be closed as far as I'm concerned.

slachtar:
I think, in the README file, there should be an explicit link to the client browser version.
https://cdn.deepstream.io/js/client/latest/ds.min.js
I'm migrating from 2.x to 5.x and the doc links to dist/... which doesn't exist.

slachtar:
Sorry @jaime-ez to bother you,
[...]

Issue #529, opened 2020-05-31
Deepstream-client have not worked in ReactNative 


codecrack3:
When I tried to connect to Deepstream and log in I get an error message. 
Does the current library does not support React Native?. Anybody can help me?, I hope to see a reply soon. Thanks

![image](https://user-images.githubusercontent.com/9525576/83345726-b6e64480-a340-11ea-82ab-dd32517fcbb0.png)

Note: I use the latest version Deepstream client and ReactNative

jaime-ez:
Hi this sould be fixed in the next release. In the meantime you can download from the latest tag [v5.1.1](https://github.com/deepstreamIO/deepstream.io-client-js/tree/v5.1.1), build the client and try it out.

jaime-ez:
Hi, this is solved and published on the latest client release v5.1.2.

Please feel free to reopen the issue if you still have problems.

yasserf:
Thanks @jaime-ez!

PR #531, opened 2020-06-28
[task] do not depend on browser or nodejs URL api [fix] export DeepstreamClient as object property in bundle

### LINKED ISSUE #515: React Native 0.59+ URLSearchParams 'Error: not implemented'

jaime-ez:
Hi @yasserf, hope everything is ok.

I gave a second thought to the react-native issue and chose not to use the URL api, neither from the browser nor nodejs, since the use case is rather simple, and coded the url string manipulation directly. This allows to use the client in web, nodejs and react-native without adding more external dependencies. This fixes #515

Also exported the client as object in order to maintain the same api in the bundle. This fixes issue #528 

Let me know what you think and I can push the package and changelog updates.

Also, should the main file in the package.json point to the bundle?

Cheers

### DIFF
--- a/src/connection/socket-factory.ts
+++ b/src/connection/socket-factory.ts
@@ -1,3 +1,4 @@
+import { Buffer } from 'buffer'
 import { parse } from '@deepstream/protobuf/dist/src/message-parser'
 import { getMessage } from '@deepstream/protobuf/dist/src/message-builder'
 import {Socket} from '../deepstream-client'
--- a/src/deepstream.ts
+++ b/src/deepstream.ts
@@ -1,3 +1,3 @@
 import { DeepstreamClient } from './deepstream-client'
 
-export = DeepstreamClient
+export { DeepstreamClient }
--- a/src/util/utils.ts
+++ b/src/util/utils.ts
@@ -1,11 +1,3 @@
-let UniformResourceLocator: any
-
-if (typeof URL === 'undefined') {
-  UniformResourceLocator = require('url').URL
-} else {
-  UniformResourceLocator = URL
-}
-
 /**
  * Compares two objects for deep (recoursive) equality
  *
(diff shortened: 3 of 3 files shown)


jaime-ez:
The explicit Buffer import made in the last commit fixes #529

yasserf:
Seems good to me. Happy to merge this in and you can commit changelog and tag directly from master. Thanks!

jaime-ez:
Ok, just did it...could you build and publish it to npm?
Best

jaime-ez:
@yasserf sorry to bother, I forgot to change the main file on package.json, in order to make it work on react-native it must be one of the bundle files, since the Buffer module is not available otherwise, so I'll put the minified one and make a new push.

One question though: since the main file on package.json is changing, the new version should be 5.1.2 or 5.2.0 ?

src/connection/socket-factory.ts
```diff
@@ -1,3 +1,4 @@
+import { Buffer } from 'buffer'
```
yasserf, line 1: why is this needed btw? Is it not polyfilled?

yasserf:
@jaime-ez you can push the tag `v5.1.2` and it should auto publish btw. CI systems yay.

I don't think changing the package.json main file makes a difference in semantic versioning since noone really uses it outside of npm (that I know of, if they do it's a bit odd and we have a small enough pool of users for it to be very unlikely).

PR #532, opened 2020-07-04
[task] implement client reset function

jaime-ez:
Hi @yasserf !

I encountered the following issue while using the v5 client:

- After closing the connection I wasn't able to connect again. For example if I'm on an app and then someone else want's to use it, I wasn't able to logout and then login with other credentials (neither the same credentials). On the web is not such an issue because you reload the page and that's it, but on react-native it's not that simple.

- Looking at the code I realised that the close method closes all services, including connection, thus making impossible to reconnect under the same instance of the client. That's why I chose to extract the main part of the client constructor to another function (called reset) which is called upon creating a new instance of the client and is also available to be called as a client method, but under this scenario it will only work if the connection status is `CLOSING` or `CLOSED`, that is only to actually reset the client after closing it.

Let me know what you think, I'm not so confident on my typescript code, but it passes all the tests and linter.

Best

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepstream/client",
-  "version": "5.1.1",
+  "version": "5.2.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepstream/client",
-  "version": "5.1.1",
+  "version": "5.2.0",
   "description": "the javascript client for deepstream.io",
   "keywords": [
     "deepstream",
@@ -8,7 +8,7 @@
     "realtime",
     "client"
   ],
-  "main": "dist/src/deepstream-client.js",
+  "main": "dist/bundle/ds.min.js",
   "types": "dist/src/deepstream-client.d.ts",
   "directories": {
     "test": "src/test"
--- a/src/deepstream-client.ts
+++ b/src/deepstream-client.ts
@@ -65,43 +65,50 @@ export {
   Options
 }
 
-class DeepstreamClient extends Emitter {
-  public event: EventHandler
-  public rpc: RPCHandler
-  public record: RecordHandler
-  public presence: PresenceHandler
+function reset (_this: DeepstreamClient): DeepstreamClient {
+  const services: Partial<Services> = {}
+  services.logger = new Logger(_this)
+  if (_this.options.nativeTimerRegistry) {
+    services.timerRegistry = new NativeTimerRegistry()
+  } else {
+    services.timerRegistry = new IntervalTimerRegistry(_this.options.intervalTimerResolution)
+  }
+  services.timeoutRegistry = new TimeoutRegistry(services as Services, _this.options)
+  services.socketFactory = _this.options.socketFactory || socketFactory
+  services.connection = new Connection(services as Services, _this.options, _this.url, _this)
+  if (_this.options.offlineEnabled) {
+    services.storage = _this.options.storage || new Storage(_this.options)
+  } else {
+    services.storage = new NoopStorage()
+  }
+  _this.services = services as Services
 
-  private services: Services
-  private options: Options
+  _this.services.connection.onLost(
+    services.timeoutRegistry.onConnectionLost.bind(services.timeoutRegistry)
+  )
+
+  _this.event = new EventHandler(_this.services, _this.options)
+  _this.rpc = new RPCHandler(_this.services, _this.options)
+  _this.record = new RecordHandler(_this.services, _this.options)
+  _this.presence = new PresenceHandler(_this.services, _this.options)
+  return _this
+}
+
+class DeepstreamClient extends Emitter {
+  public event!: EventHandler
+  public rpc!: RPCHandler
+  public record!: RecordHandler
+  public presence!: PresenceHandler
+
+  public services!: Services
+  public options: Options
+  public url: string
 
   constructor (url: string, options: Partial<Options> = {}) {
     super()
     this.options = { ...DefaultOptions, ...options } as Options
-    const services: Partial<Services> = {}
-    services.logger = new Logger(this)
-    if (options.nativeTimerRegistry) {
-      services.timerRegistry = new NativeTimerRegistry()
-    } else {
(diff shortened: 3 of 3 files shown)


yasserf:
So there’s a pause and resume method, and reconnection logic already exists.

As such I’m not fully following. I get the issue with the Auth credentials but I think the issue your facing is more around your app design rather than deepstream tbh.

If you logout your app should be destroyed in react native (everything unmounts). If your carrying state between different users it’s sort of a security concern (worst case).

jaime-ez:
Yes, it is not related to reconnection logic. I'll try to clarify with an example.

If you run this index.html:
```
<!DOCTYPE html>
<html lang="en" dir="ltr">
  <head>
    <meta charset="utf-8">
    <script src="https://unpkg.com/@deepstream/client@5.1.1/dist/bundle/ds.min.js"></script>
  </head>
  <body>
    <script type="text/javascript">
      window.onload = function () {
        const client = new DeepstreamClient.DeepstreamClient('localhost:6020')
        client.login({username: 'peter', password: 'sesame'}, (success, data) => {
          console.log('SUCCESS1', success)
          client.close()
          setTimeout(function () {
            console.log('relogin')
            client.login({username: 'peter', password: 'sesame'}, (success, data) => {
              console.log('SUCCESS2', success)
            })
          }, 1);
        })
        client.on('connectionStateChanged', (state) => {
          console.log('state', state)
        })
      }
    </script>
  </body>
</html>
```

You will see that the second login does not trigger. `SUCCESS2` is never printed since the client is in closed state.

I think this behaviour is ok, but I would like the option to be able to login again on the same scope if I want to (as was possible with the v2 client).

That's why I implemented the reset method that allows you to do this:

```
<!DOCTYPE html>
<html lang="en" dir="ltr">
  <head>
    <meta charset="utf-8">
    <script src="http://localhost:8081/ds.min.js"></script>
    <!-- <script src="https://unpkg.com/@deepstream/client@5.1.1/dist/bundle/ds.min.js"></script> -->
  </head>
  <body>
    <script type="text/javascript">
      window.onload = function () {
        const client = new DeepstreamClient.DeepstreamClient('localhost:6020')
        client.login({username: 'peter', password: 'sesame'}, (success, data) => {
          console.log('SUCCESS1', success)
          client.close()
          setTimeout(function () {
            console.log('relogin')
            client.reset()
            client.login({username: 'peter', password: 'sesame'}, (success, data) => {
              console.log('SUCCESS2', success)
            })
          }, 1);
        })
        client.on('connectionStateChanged', (state) => {
          console.log('state', state)
        })
      }
    </script>
  </body>
</html>
```

Now after calling reset you can connect again if you want to.

My use case arises from 2 situations:

- I'm using react-redux, so I initialise the client on the store and pass it along to components via the store middleware, so I can have the same client available on all action creators. Thus after logout, the store is cleaned but not unmounted, so I don't reload the client per se. Thus I need to be able to relogin without reloading the client.

- I have some users that can access multiple customers data, but for compliance I let them do so one at the time, thus when they login they send an activeCompany flag with their credentials and thus they have access to that company data via valve permissions check on that flag. That's why I need them to be able to logout-login without unmounting the whole app.

This reset method was the fastest/first way I came up for solving the problem without changing the current behaviour. I think it is a sane option to have, independent of the app design.

yasserf:
So the issue I'm seeing with the reset method that I can see is that it essentially rewrites all of the client state. So you lose all references to current subscriptions (invalidating anything you already have) and essentially just using the deepstream instance as a very shallow wrapper.

So rather than using a factory/singleton pattern (which is what I would argue is a bit more sane) your now making a single deepstream instance responsible for what is essentially replacing it's entire internal logic.

It would probably easier in the react world to use a context (to retrieve the client) or in the redux world to use a singleton pattern where you can do (deepstream().event.subscribe).

In client v2 is just recreated the socket after changing auth params (same as .resume):
https://github.com/deepstreamIO/deepstream.io-client-js/blob/v2/src/message/connection.js#L82

In v4 the same thing is there but it's commented out:
https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/connection/connection.ts#L227

Seems like it was commented out a few years ago by one of the developers due and I just never commented it back in.

jaime-ez:
As always, thanks for sharing your wisdom!

I went for the singleton pattern, works as expected, since it wasn't necessary on the older client I hadn't thought about it.

Uncommenting this code:
> In v4 the same thing is there but it's commented out:
https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/connection/connection.ts#L227

The condition is never met since the endpoint is defined, changing the condition to this:
```
    if (this.stateMachine.state === CONNECTION_STATE.CLOSED || this.stateMachine.state === CONNECTION_STATE.CLOSING) {
      this.createEndpoint()
      return
    }
``` 
throws invalid state transition errors when trying to relogin.

So in summary I'll close this pull request and try to make one to the website giving the warning that once the connection is closed it can not be reopened on the same instance.

And one last point, did you saw my last comment [here](https://github.com/deepstreamIO/deepstream.io-client-js/pull/531#issuecomment-653187629)? 

Thanks!

yasserf:
Happy to hear it works well for you

I think if I was to raise a PR to allow a user to change their auth credentials without invalidating what I would do is:

```
.pause()
.updateAuth() // new API that replaces auth credentials and nothing else
.resume()
```

or

```

```
.updateAuth() // new API that replaces auth credentials and nothing else
```

That way you can either predict auth credentials changing (like a JWT token) without having to kill the connection or update it on a INVALID_AUTH event.

I generally like the connection being completely dead at the end at the end tbh, because it forces things to clean up if you surpass the maximum amount of invalid login events (including dumping you local storage if configured that way).

yasserf:
Interestingly in polymer tools and older (like 4 years ago) angular I didn't use singletons:

https://github.com/deepstreamIO/deepstream.io-tools-polymer/blob/master/ds-connection-behaviour.html
https://github.com/deepstreamIO/ds-demo-simple-app-ng/blob/master/src/start-ng.js

But my reasoning in React and Angular2+ for some reason changed. So I can see the confusion, I myself am confused haha

jaime-ez:
Yeah it can be a bit confusing, at first I thought it would be logic to allow a relogin, but know I see it's cleaner the way it is.

I'll leave the suggestion you make for changing auth credentials as an issue to be handled in the future if required. :+1:

PR #538, opened 2020-08-12
[fix] discard all references of the record

jaime-ez:
I encountered the following issue: calling record.discard() for a record that has been retrieved multiple times does not actually discard all the record instances nor unsubscribes from them. Example:

```
const { Deepstream } = require('@deepstream/server')
const { DeepstreamClient } = require('@deepstream/client')

const run = async () => {
  const server = new Deepstream({})
  server.start()

  server.on('error', (e) => {
    console.error(e)
  })

  server.on('started', () => {
    console.log('Started')
    const client = new DeepstreamClient('localhost:6020')
    client.login(() => {
      client.record.getRecord('user').whenReady((record) => {
       // log data changes
        record.subscribe((data) => {
          console.log('subscribed data', data)
        })
       // this should be logged
        record.set('one', 1)

       // this is called after the timeout a few lines below
        setTimeout(function () {
          console.log('end')
          // this should not trigger the subscription log
          record.set('three', 3)
        }, 10 * 1000)
      })

      setTimeout(function () {
        client.record.getRecord('user').whenReady((record) => {
         // this should be logged
          record.set('two', 2)
          // record.unsubscribe()
          record.discard()
        })
      }, 5 * 1000)
    })
  })
}

run()
```
After calling record.discard() the subscribed data is still logged.

If you replace discard for unsubscribe, the subscription is actually removed and there will be no final log, although the record will still be present on the client.

The commit fixes this behaviour, so all instances of the record are unsubscribed and removed from the client. I imagine this is the expected behaviour.

Let me know what you think in order to push version bump and changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+## [5.1.5] - 2020.08.18
+
+### Fix  
+
+- Discard all references of a record when calling `.discard()`
+- Remove subscriptions from record reference when calling `.unsubscribe()`
+
+## [5.1.4] - 2020.07.29
+
+### Chore  
+
+- Bump dependencies
+
 ## [5.1.3] - 2020.07.14
 
 ### Fix
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepstream/client",
-  "version": "5.1.4",
+  "version": "5.1.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepstream/client",
-  "version": "5.1.4",
+  "version": "5.1.5",
   "description": "the javascript client for deepstream.io",
   "keywords": [
     "deepstream",
--- a/src/record/record-core.ts
+++ b/src/record/record-core.ts
@@ -117,6 +117,13 @@ export class RecordCore<Context = null> extends Emitter {
   })
 }
 
+  /**
+  * Removes scope from emitter
+  */
+  public removeContext (ref: any): void {
+    this.emitter.removeContext(ref)
+  }
+
   private onDirtyServiceLoaded () {
     this.services.storage.get(this.name, (recordName, version, data) => {
       this.services.connection.onReestablished(this.onConnectionReestablished)
--- a/src/record/record.ts
+++ b/src/record/record.ts
@@ -101,12 +101,14 @@ export class Record extends Emitter  {
 
     public unsubscribe (path: string, callback: (data: any) => void): void {
         const parameters = utils.normalizeArguments(arguments)
-        this.subscriptions = this.subscriptions.filter(subscription => {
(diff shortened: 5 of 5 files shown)


jaime-ez:
I'm gonna try another approach. The unsubscribe functionality shouldn't unsuscribe from all record instances, just the one it was called from. The discard function should be kept as it was and a new discardAll method could be implemented. I'll try to update the code soon.

jaime-ez:
Finally I left it as it is. The behaviour works as expected. Iterating through all record instances is necessary in order to transition to the unsubscribe state. I added the removeContext method becasue the context was not being removed properly from the emitter.

yasserf:
late reply on this. Came back to europe and haven't done much work on anything really while trying to figure things out (and avoiding wasps).

This fix is actually a regression I'm afraid, calling discard should only be scoped on to the instance you want and not everything. If you want to force a discard on all records with the same name we would need add a seperate API on the record handler and not on the record itself.

```
const x = getRecord('a')
x.subscribe(console.log)

const y = getRecord('a')
y.subscribe(console.log)

x.discard()
```

That should remove one count and all of the subscriptions on x, but y keeps the record alive (not discarding it) and it should only log stuff once.

jaime-ez:
Hi, glad you avoided wasps!

1. Go ahead with rolling back.

2. as you can see from my previous monologue, I wasn't sure of the expected behaviour. I decided this way because:

- docs say: `discard()`: _Removes all change listeners and notifies the server that client no longer wants updates for this record if your application no longer requires the record._ I interpreted that as the whole client is no longer interested on the record, not just on the particular instance.

- It is the current behaviour that `unsubscribe()` has, as can be seen on the example I posted on the previous comment, calling `unsubscribe()` affects all the record instances. Which is what lead me to this issue in the first place.

I propose then:

1. `unsubscribe()` and `discard()` should only affect the record instance they are called upon.

2. create new methods on record core `discardAll()` and `unsubscribeAll()` that implement the current behaviour this commit proposed.

3. update docs

What do you think?

yasserf:
i think that sounds great.

> docs say: discard(): Removes all change listeners and notifies the server that client no longer wants updates for this record if your application no longer requires the record. I interpreted that as the whole client is no longer interested on the record, not just on the particular instance.

Yeah this is more v3 of the record. Although, in v3 it just used a number counter (instead of scope) so you would do something like this:

```
const x = getRecord('a')
const y = getRecord('a')
x.discard()
x.discard() // although x is discarded already x and y are the exact same object
```

> create new methods on record core discardAll() and unsubscribeAll() that implement the current behaviour this commit proposed.

So discardAll is more of a bandaid than a solution. Ideally each record would be responsible for it's own lifecycle. Discard all could be a solution but it might lead to bad practice. 

It looks like the issue is with the scope when doing a subscribe

The emitter expects: `eventName, callback, context` in order for clearContext to work. The record just subscribed with eventName and callback (the context defaults to that the record-core). Seems like the fix might be to ensure the context is always provided correctly from list/record/anonymous-record

PR #539, opened 2020-08-24
[fix] discard all instances of list

jaime-ez:
this replicates the last change to record discard

### DIFF
--- a/src/record/list.ts
+++ b/src/record/list.ts
@@ -47,14 +47,20 @@ export class List extends Emitter {
     }
 
     public discard (): void {
-      this.destroy()
-      this.record.removeReference(this)
+      for (const item of this.record.references) {
+        for (let i = 0; i < item.subscriptions.length; i++) {
+          this.record.unsubscribe(item.subscriptions[i])
+        }
+        item.wrappedFunctions.clear()
+        this.record.removeReference(item)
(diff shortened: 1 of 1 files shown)


yasserf:
This breaks the encapsulation concept in the client, which is that the record core is source of truth and the record/anonymous record/list are just references.

This fix means if a client does this:

```
const x = getList('a')
const y = getList('a')
x.discard()
```

Then both x and y (and any other reference to list 'a') would be discarded which would break complex applications quite easily.

yasserf (changes requested):
This uses the score of the record (via this.record.references) which can cause issues by affecting other records.

I'm happy to have a. conversation about this to understand the issue better. If this was merged for records we might want to roll that back as well in favour of whatever solution we come up with.

Thanks for raising, I know some of the arch concepts are a bit odd at first glance.

Brief reason it exists as a wrapped (if you compare it to V3 and before):

In V3 you would get a record which and it would essentially return what we have as record-core in V4. This means you HAVE to subscribe and unsubscribe to everything, otherwise you have a memory leak.

In V4 you have a thin wrapper which allows you to call discard on a list or record which will remove all local references but not affect any other instances of the same record/list. So cleanup code and leaks is much simpler.

PR #541, opened 2020-09-02
fix: provide context for record subscriptions

jaime-ez:
This should fix the problems with unsubscribing from a record and keeping other record's subscriptions active.
I had to put a couple more conditions on the emitter file open for review.

The previous pull request was closed to the history rewrite (a bad choice)

### DIFF
--- a/src/record/list.ts
+++ b/src/record/list.ts
@@ -196,7 +196,7 @@ export class List extends Emitter {
         parameters.callback = listCallback
 
         this.subscriptions.push(parameters)
-        this.record.subscribe(parameters)
+        this.record.subscribe(parameters, this)
     }
 
     /**
@@ -212,8 +212,13 @@ export class List extends Emitter {
 
         const listenCallback = this.wrappedFunctions.get(parameters.callback)
         parameters.callback = listenCallback as (data: any) => void
-        this.record.unsubscribe(parameters)
         this.wrappedFunctions.delete(parameters.callback)
+        this.subscriptions = this.subscriptions.filter((subscription: any) => {
+          if (parameters.callback && parameters.callback !== subscription.callback) return true
+          return false
+        })
+
+        this.record.unsubscribe(parameters, this)
     }
 
 /**
@@ -340,7 +345,7 @@ private applyUpdate  (message: RecordMessage) {
 
   private destroy () {
     for (let i = 0; i < this.subscriptions.length; i++) {
-      this.record.unsubscribe(this.subscriptions[i])
+      this.record.unsubscribe(this.subscriptions[i], this)
     }
     this.wrappedFunctions.clear()
     this.record.removeContext(this)
--- a/src/record/record-core.ts
+++ b/src/record/record-core.ts
@@ -117,13 +117,6 @@ export class RecordCore<Context = null> extends Emitter {
   })
 }
 
-  /**
-  * Removes scope from emitter
-  */
-  public removeContext (ref: any): void {
-    this.emitter.removeContext(ref)
-  }
-
   private onDirtyServiceLoaded () {
     this.services.storage.get(this.name, (recordName, version, data) => {
       this.services.connection.onReestablished(this.onConnectionReestablished)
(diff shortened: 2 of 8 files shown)


src/record/list.ts
```diff
@@ -212,8 +212,13 @@ export class List extends Emitter {
 
         const listenCallback = this.wrappedFunctions.get(parameters.callback)
         parameters.callback = listenCallback as (data: any) => void
-        this.record.unsubscribe(parameters)
         this.wrappedFunctions.delete(parameters.callback)
+        this.subscriptions = this.subscriptions.filter((subscription: any) => {
+          if (parameters.callback && parameters.callback !== subscription.callback) return true
```
yasserf, line 217: tiny syntax sugar, but you can just do

this.subscriptions.filter((subscription: any) => parameters.callback !== subscription.callback)

I'm assuming callback always exists, since otherwise its broke

src/util/emitter.ts
```diff
@@ -56,6 +56,17 @@ export class Emitter {
...
     for (let i = 0; i < callbacks.length; i++) {
       const { fn: cb, scope: context } = callbacks[i]
+
+      // handle unsubscribing from all callbacks for a given record path
+      if (event !== '' && fn === undefined && scope === context) {
+        callbacks.splice(i, 1)
+      }
+
+      // handle unsubscribing from all callbacks for a list
+      if (event === '' && fn === undefined && scope === context && scope.constructor.name === 'List') {
```
yasserf, line 66: whats the difference between a record and list in this context? 

Also an emitter should not know about data types, since it's used for all pubsub client side

src/record/record.ts
```diff
@@ -96,31 +96,26 @@ export class Record extends Emitter  {
...
-        for (const item of this.record.references) {
-          item.subscriptions = item.subscriptions.filter((subscription: any) => {
-            if (!parameters.callback && (subscription.path === parameters.path)) return false
-            if (parameters.callback && (subscription.path === parameters.path && subscription.callback === parameters.callback)) return false
-            if (parameters.callback && subscription.callback === parameters.callback) return false
-            return true
-          })
-        }
+        this.subscriptions = this.subscriptions.filter((subscription: any) => {
+          if (!parameters.callback && (subscription.path === parameters.path)) return false
```
yasserf, line 105: i know this is older code, but when would we not have a callback? I feel like that might be broken code

src/record/record-core.ts
```diff
@@ -117,13 +117,6 @@ export class RecordCore<Context = null> extends Emitter {
   })
 }
 
-  /**
-  * Removes scope from emitter
-  */
-  public removeContext (ref: any): void {
```
yasserf, line 123: this is references in other places, is it meant to be deleted?

yasserf (changes requested):
Some small requests, thanks for raising!

src/record/record-core.ts
jaime-ez: I added that method on the latest realease, it wasn't present on previous versions. Since I did a rollback it got removed again. I think when it is called in other places the prototype chain calls the removeContext method from Emitter, since Record Core is an extension of it. We might keep it though on record core for clarity?

src/record/record.ts
jaime-ez: when you do `unsubscribe('path')`

src/util/emitter.ts
jaime-ez: Yes I'm not very happy with this. 

I digged into it again and found a bug that affects this logic. Will make a new commit to further discuss it

src/record/list.ts
jaime-ez: Nope callback doesn't always exist. You can do `list.unsubscribe(callback)` to cancel a specif callback or `list.unsubscribe()` to get rid of all subscriptions, that's why the extra step. Because otherwise `undefined !== function` is always true

jaime-ez:
Ok, just pushed a (hopefully) fix to the emitter logic. The emiter Map was being modified directly by using a for loop and `callback.splice()` which mutated the array, and thus the last element wasn't evaluated. For example a callback array of length 2 where the callback[0] element called callback.splice(), would never enter into the for loop for callback[1] since the length of the array was reduced and thus the condition never met. I changed it to use a filter and then if the resulting array is not empy, we set the callbacks for the event key again.

yasserf:
LGTM

yasserf (approved):

PR #542, opened 2020-09-10
fix: remove callback

jaime-ez:
Adding a notifier when offline called the callback inmediately, but it wasn't removed from the requests object, adn therefore future calls would not send the message


yasserf:
Awesome!

PR #544, opened 2020-09-17
change bundle file for react-native

jaime-ez:
When building the react native app for production, the minified client file does not compile correctly for react-native's javascript engine. Therefore the non minified bundle must be used and if a minifier is used for building the react-native app, some configuration options must be set that will be described in the website docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [5.1.8] - 2020.09.17
+
+### Fix  
+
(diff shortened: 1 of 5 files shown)


nikoTM:
Any chance this is impacting Nativescript as well?

jaime-ez:
Possible, if Buffer is not defined in the nativescript engine. Try installing as documented for react-native, let me know if it works in order to add it to the docs

PR #549, opened 2020-11-28
fix: send ping message based on lastSentMessage timestamp

msmarks:
On the client side, the original logic is that its checks if a message wasn't received for a `heartbeatInterval` and decide to send a ping. But if the client only just subscribes to the change of server data and not publish any message, the time interval of server data change is always less than `heartBeatInterval`, the client will not send any Ping, so the server will not receive any data from the client, and the client will be considered disconnected.

Change `lastRecievedMessageTimestamp` to `lastSentMessageTimestamp`, which can ensure that the client sends a message to the server at least once in a `heartbeatInterval `.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [5.1.11] - 2020.12.1
+
+### Fix
+
+- Send heartbeat packets according to `lastSentMessageTimestamp`, not `lastRecievedMessageTimestamp`.
+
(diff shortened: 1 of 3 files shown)


yasserf:
This looks good to me thanks! If you can add something to the changelog I'll get it merged published ASAP

msmarks:
Thank you. The change log has been added.

yasserf:
Amazing thanks

Issue #551, opened 2020-12-10
ACCEPT_TIMEOUT weired behavior


slachtar:
I'm working on a slow server (sandbox for tests) and set the following values on server config
```rpc:
  # Timeout for client RPC acknowledgement
  ackTimeout: 180000
  # Timeout for actual RPC provider response
  responseTimeout: 180000
```

I do use default acceptTimout on client side (which is 6000 I think).
When I make an rpc call twice for the same function, the second call raises an ACCEPT_TIMEOUT error and the rpc callback is called which is fine for me.

The problem is that after some time (few seconds), the rpc callback is called again but now with result returned from backend.
Is this the expected behavior?

When this happens, I can continue making other rpc calls, but cannot subscribe to records anymore!

jaime-ez:
Hi, I was just testing a similar problem and came to see if it had been raised before! I'm seeing the behaviour with `rpcResponseTimeout`:

If I declare on the client options a value for rpcResponseTimeout that is less than the server default or declared, I get a RESPONSE_TIMEOUT error, and then the response from the server. This should not happen in my opinion becasue it leads to bugs: you handle the error but then the success response is called anyway.

I think the solution is to add a isComplete flag on rpc.ts file, I'll make a pull with the proposed changes

slachtar:
Yes, that was exactly the behavior with the accept timeout too. After that
error, the deepstream client stops working properly.

It will be great if you could fix it.

Thank you,

jaime-ez:
Well, the problem has been eliminated on version 5.2.0 (to be released). The client no longer controls rpc timeouts. I'll update the website soon

slachtar:
Thank you, nice work.

PR #553, opened 2020-12-11
fix: add isComplete flag to prevent callback being called more than once

jaime-ez:
Possible fix for #551

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [5.2.0] - 2020.12.29
+
+### Breaking Change
+
+Remove rpcAcceptTimeout and rpcResponseTimeout from client logic. These timeouts will be handled server side only.
+
 ## [5.1.12] - 2020.12.4
 
 ### Fix
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepstream/client",
-  "version": "5.1.11",
+  "version": "5.2.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepstream/client",
-  "version": "5.1.12",
+  "version": "5.2.0",
   "description": "the javascript client for deepstream.io",
   "keywords": [
     "deepstream",
--- a/src/client-options.ts
+++ b/src/client-options.ts
@@ -30,14 +30,6 @@ export interface  Options {
      * Default: 5
      */
     maxReconnectAttempts: number
-    /** The number of milliseconds after which a RPC will error if no Ack-message has been received.
-     * Default: 6000
-     */
-    rpcAcceptTimeout: number
-    /** The number of milliseconds after which a RPC will error if no response-message has been received.
-     * Default: 10000
-     */
-    rpcResponseTimeout: number
     /** The number of milliseconds that can pass after providing/unproviding a RPC or subscribing/unsubscribing/listening to a record or event before an error is thrown.
      * Default: 2000
      */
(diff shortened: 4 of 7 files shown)


yasserf:
If the server is sending a response timeout that shouldn’t affect the client sending a response afterwards. Seems more like the server should respond with an unknown rpc correlation id

jaime-ez:
The server is not sending a response timeout,  if the client has a timeout less than the server one, the client locally calls the error [here](https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/rpc/rpc.ts#L85) without sending a message to the server, which I think is fine becasue it is a local decision to define that the timeout has been exceded, but it wasn't declaring locally that the rpc has been finished, thus the "double callback" with the incoming server message. I propose that adding this flag will cause that the incoming server message, if sent, will be ignored.

yasserf:
Gotcha.

I would think it makes more sense for the client not to do the response_timeout, and just let the server handle it. The issue is primarily around users configurations being invalid.

Silencing errors using guards is a tricky decision, because on one side we are making it work 'as expected', but on the other side we still have this RPC response being sent up all the way to the client which is invalid.

I would say the two things we can do is:
1) Send the required server config to the client (like heartbeat/timeouts). The client can then validate it.
2) Just give better defaults to the client, where the response time is much higher than the server.

The RPCs should all be marked invalid if the connection to the server is lost, so there shouldn't be any reason to have timeouts client side that change state, we usually only use them to print warnings.

jaime-ez:
So basically the client should'nt have any rpc timeout options (neither accept nor response) and just let the server handle it? I don't see the point on validating configs between server and client if we can not override the server values client-side.

yasserf:
It isn't to validate config client side, its to set the config entirely for certain things on the client.

For example people are running into issues with the heartbeat on the server and client not matching.

In terms of the RPC timeout, it really doesn't matter much, the whole concepts of timeouts that are baked into deepstream since v0.0 don't do much other than log warnings. As long as the connection state is properly maintained (by using correct heartbeat logic) then the user will either get the timeout from the server or the connection will be considered lost and will trigger error callbacks.

I would say we just delete timeout issues on the client as long as the connection works.

jaime-ez:
> For example people are running into issues with the heartbeat on the server and client not matching.

Have those issues been solved?

> I would say we just delete timeout issues on the client as long as the connection works.

Yes I agree, less config is better. I'll update with changes reflecting this.

yasserf:
The accept timeout is used by the server to indicate a backend rpc provider hasn't confirmed it is willing to satisfy the request, which the server takes as a sign to try another one.

Generally I would argue that sort of logic is a bit extreme, if you make a request and it fails then the server should just pass on that it failed to respond, to avoid having the same thing processed twice. We don't tell the clients to abort either so the response goes into the void.

Like I said, timeouts are pretty excessively implemented, should have dropped more of them when doing the v4 refactor. We can use timeouts to warn that the server is being slow but it should just fail hard.

> Have those issues been solved?

I believe they have gotten better, but there was a PR open somewhere than still needs to be merged.

jaime-ez:
Pull updated:
- removed rpc timeouts from client, therefore the accept and response timeouts are handled server side only
- I kept the rpc.accept method just in case, I think it could be safely removed from the client
- this new version should be updated to 5.2.0 in order to reflect that this update might break something for users that rely on client timeouts.
- Docs will be updated pending approval

Comments?

yasserf:
This looks good to me. Do we have any tests to make sure that when connection drops the RPCs fails?

jaime-ez:
> This looks good to me. Do we have any tests to make sure that when connection drops the RPCs fails?

apparently yes: https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/rpc/rpc-handler.spec.ts#L553

yasserf:
Apologies for being nitpicky in advance 🙈

The one you linked to tests what happens when the response is an error (so it goes to the provider, errors because of the provider, goes to deepstream and then back to the requestor).

The test that covers a network loss is here:

https://github.com/deepstreamIO/deepstream.io-client-js/blob/master/src/rpc/rpc-handler.spec.ts#L579

jaime-ez:
yes I meant to reference that line...wrong copy/paste :+1:

PR #558, opened 2021-02-24
Error when running under vue3 / vite

slumtrimpet:
"global" is not valid in client-side code.  Here is one proposed patch that gets things working, but possibly a different refactor makes more sense.

### DIFF
--- a/src/connection/socket-factory.ts
+++ b/src/connection/socket-factory.ts
@@ -3,7 +3,7 @@ import { getMessage } from '@deepstream/protobuf/dist/src/message-builder'
 import {Socket} from '../deepstream-client'
 import { JSONObject, TOPIC, Message, CONNECTION_ACTION } from '../constants'
 
-const global = window
+const global = global || window
 const BrowserWebsocket = (global.WebSocket || global.MozWebSocket) as any
 
 export type SocketFactory = (url: string, options: JSONObject, heartBeatInterval: number) => Socket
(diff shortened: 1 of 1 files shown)


slumtrimpet:
Possibly a fix to get the build checks working could be:

```
var global = global || window;
```

But, the access of 'global' here is still hokey.

jaime-ez:
Hi, you should use the webpack bundle that handles this automatically. Check your dev setup, raise an issue if you have any questions.

slumtrimpet:
> Hi, you should use the webpack bundle that handles this automatically. Check your dev setup, raise an issue if you have any questions.


The issue is VUE3 standard install doesn't use webpack anymore, it uses VITE.
https://v3.vuejs.org/guide/installation.html#vite

slumtrimpet:
> Ok, so you need to use the bundle file. See the readme and try to install the client as if you where on react native.

Gotcha!  It's a bit of a WIP hack at the moment, but the following seems to make Vue3 + Vite + Typescript happy and I'm continuing with testing the platform.

```
import { DeepstreamClient } from "@deepstream/client";
import { DeepstreamClient as DeepstreamDistClient } from "@deepstream/client/dist/bundle/ds.js";
const client: DeepstreamClient = new DeepstreamDistClient("localhost:6020");
client.login();
```

Thanks for your help on this!

jaime-ez:
great, please feel free to make pulls to the readme in order to alert other vite users

PR #570, opened 2022-01-08
Fix: Version conflict when receiving 'PATCH' action causes record to go into 'MERGING' state indefinitely

Vortex375:
I had a producer which frequently updated a record using the "patch" operation (i.e. setting only one property of the record). I noticed that occasionally my subscribers stopped handling updates.

Through debugging I found that when the subscriber stopped handling updates, the record would go into "MERGING" state and then never leave that state. After further debugging, I think that the cause for this is that the record issues a "read" request but forgets to register a callback. Thus, the read response is not handled and the record never leaves the "MERGING" state.

Properly registering the callback for the read request seems to fix the problem I was experiencing.

### DIFF
--- a/src/record/record-core.ts
+++ b/src/record/record-core.ts
@@ -719,6 +719,7 @@ export class RecordCore<Context = null> extends Emitter {
         * the full state of the record
         **/
         this.sendRead()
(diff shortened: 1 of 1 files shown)


jaime-ez:
Hi, thanks for the pull request! Your fix makes sense :+1:

